PR #5, opened 2021-09-22
feat: update bundling

SamuelAlev:


oliverschwendener (approved):
Tested it and it works as expected 💪 Thanks alot for taking care of this!

PR #123, opened 2021-12-15
feat: use webpack as experimental bundler

triggertoo:


src/index.ts
```diff
@@ -1,16 +1,26 @@
...
-const parseArgs = minimist(process.argv.slice(2));
+const experimentalOptionName = "experimental";
+const options = buildOptions({
+    [experimentalOptionName]: {
+        type: "boolean",
+        alias: "e",
+        default: false,
+    },
+});
+const parseArgs = minimist(process.argv.slice(2), options);
```
SamuelAlev, line 23: Feel ashamed I haven't seen that before 🙈

src/utils/compile.ts
```diff
@@ -106,3 +131,119 @@ export const compile = async (
...
+): Promise<void> => {
+    const compiler = webpack({
+        mode: options.env?.NODE_ENV === "development" ? "development" : "production",
+        context: projectPath,
+        externals: {
+            react: "React",
+            "react-dom": "ReactDOM",
+        },
+        entry: {
+            "./index.tsx": getVirtualEntry(entryFileNames),
```
SamuelAlev, line 169:
```suggestion
            "./_virtual/index.tsx": getVirtualEntry(entryFileNames),
```
To be extra explicit about it

package.json
```diff
@@ -25,19 +25,31 @@
...
         "@rollup/plugin-replace": "3.0.0",
+        "babel-loader": "^8.2.3",
+        "css-loader": "^6.5.1",
         "esbuild": "0.14.1",
         "fastify": "3.24.1",
         "fastify-cors": "6.0.2",
         "fastify-static": "4.5.0",
         "fastify-websocket": "4.0.0",
+        "minimist": "^1.2.5",
+        "minimist-options": "^4.1.0",
```
SamuelAlev, line 45: Is it required to be in `dependencies`?
As everything is bundled it should not

triggertoo: i guess it's a decision of the package scope. since this one is about to provide the tooling, it cannot run without having it, so it's a dependency imho

PR #258, opened 2022-10-05
fix: change host

SamCreasey:


triggertoo (approved):

SamuelAlev (changes requested):
We need another approach to this, it is a security issue now.

SamuelAlev:
rebased on latest main in #265

PR #259, opened 2022-10-11
feat: add required files for linting

SamCreasey:
Since users can upload code with the "--noVerify" flag, we have to check the code quality for public blocks during the review process. Not having a prettierrc, tsconfig or eslintrc file will cause the compilation and quality checks to fail so it makes sense that these files must exist.
A .gitignore is not technically necessary for anything, but it was throwing an error and causing the review process to exit early if it was not found, so we should allow the user to upload without a gitignore

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -6,6 +6,7 @@ import { Headers } from 'node-fetch';
 import open from 'open';
 import { join } from 'path';
 import CompilationFailedError from '../errors/CompilationFailedError';
+import FileNotFoundError from '../errors/FileNotFoundError';
 import { compile } from '../utils/compiler';
 import { Configuration } from '../utils/configuration';
 import { readFileAsBase64, readFileLinesAsArray } from '../utils/file';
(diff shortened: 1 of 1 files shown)


src/commands/deploy.ts
```diff
@@ -69,15 +78,21 @@ export const createDeployment = async (
             const BUILD_FILE_BLOCK_LIST = ['**/*.*.map'];
             const buildFilesToIgnore = BUILD_FILE_BLOCK_LIST.map((path) => join(projectPath, path));
 
+            const SOURCE_FILES_REQUIRED_LIST = ['.prettierrc', '.eslintrc.js', 'tsconfig.json'];
```
SamuelAlev, line 81: Prettier config can be defined as followed:
```
    A "prettier" key in your package.json file.
    A .prettierrc file written in JSON or YAML.
    A .prettierrc.json, .prettierrc.yml, .prettierrc.yaml, or .prettierrc.json5 file.
    A .prettierrc.js, .prettierrc.cjs, prettier.config.js, or prettier.config.cjs file that exports an object using module.exports.
    A .prettierrc.toml file.
```
And it is not a viable solution if the dev want to push the code without prettier (I don't use prettier on personal project, just eslint), some others just don't use anything and we should most likely respect that

SamCreasey: @SamuelAlev Interesting. In this instance we should remove prettier here, but the regex of allowed files would need to be extended to allow for .json5, .cjs, .yaml and .toml files too. If the user modifies their eslint rules but we don't upload the prettier file that they use it could cause linting errors on our end when performing the review

PR #261, opened 2022-10-12
refactor: enhance CLI

SamuelAlev:
- replaced `minimist` (no longer maintained, repo got deleted) by `cac` (used by `vite`)
~~replaced `node-fetch` with `undici` (`undici` is packed with Node 18, so we will be able to drop it from `package.json` soon)~~ 
Support for `fetch` is not yet existant by most HTTP mock libraries (`msw`, `nock`, ...)
- replaced `colors` with `picocolors` (`colors` had a security issue recently, `picocolors` is alos used by `vite`)
- added `prompts` for a ✨ nicer ✨ experience when creating a project
- prefixed node builtins with `node:` as it is recommended
- changed CLI to esmodule

TODO:
- [x] Fix tests in `logout.spec.ts`, `httpClient.spec.ts` and `user.spec.ts`


src/index.ts
```diff
@@ -1,139 +1,115 @@
...
+cli.command('logout', 'log out of an instance').action(logoutUser);
 
-        case 'logout':
-            logoutUser();
-            exit(1);
+/**
+ * @deprecated `block serve` and `theme serve` will be removed in version 4.0 in favour of `serve`
+ */
+for (const appType of ['block', 'theme']) {
+    cli.command(`[root] ${appType} serve`, `serve the ${appType} locally`)
```
triggertoo, line 51: Maybe mention it here as well that it's being deprecated in v4.0, so the user can immediately see it?

```suggestion
    cli.command(`[root] ${appType} serve`, `[deprecated: use 'serve' instead] serve the ${appType} locally`)
```

src/index.ts
```diff
@@ -1,139 +1,115 @@
...
+cli.command('serve', 'serve the app locally')
+    .alias('dev')
+    .option('-e, --entryPath, --entry-path <entryPath>', '[string] path to the entry file', {
+        default: join('src', 'index.ts'),
+    })
+    .option('--port <port>', '[number] specify port', {
+        default: process.env.PORT || 5600,
+    })
+    .action(async (_args, options) => {
+        await createDevelopmentServer(options.entryPath, options.port);
```
triggertoo, line 73: this fails fore me when calling it without the `entryPath` parameter:
Screenshot 2022-10-24 at 09 08 24

src/index.ts
```diff
@@ -1,139 +1,115 @@
...
+cli.command('create [appName]', 'create a new marketplace app').action(async (appName: string) => {
+    let promptedName: string | null = null;
+
+    if (!appName) {
+        const { value } = await prompts({
+            type: 'text',
+            name: 'value',
+            message: 'App Name',
+            initial: 'my-frontify-app',
+            validate: (value: string) => (value === '' ? 'You need to enter an app name.' : true),
```
triggertoo, line 99:
```suggestion
            validate: (value: string) => (value.trim() === '' ? 'You need to enter an app name.' : true),
```

triggertoo (approved):

PR #265, opened 2022-10-31
feat: allow external ips to access server

SamuelAlev:
- Allow any IPs to access the dev server with `--allowExternal`
- Remove the `fs.strict = false` as [it has been fixed ](https://github.com/vitejs/vite/issues/2820)
- Remove unused type `Settings` as it has been moved to `@frontify/guideline-blocks-settings`


SamCreasey (approved):
Nice addition, good to see the security issue fixed too 👍

PR #301, opened 2022-12-14
fix: missing update of page in category in event

GoranRibic:


packages/app-bridge/src/react/useDocumentPages.ts
```diff
@@ -33,29 +33,22 @@ export const useDocumentPages = (appBridge: AppBridgeTheme, documentId: number)
             setDocumentPages((previousState) => {
                 const isInitial = previousState === null;
 
+                const item = event.documentPageOrDocumentCategory as DocumentPage | DocumentCategory;
```
SamuelAlev, line 36: It is not an `item` but `documentPageOrDocumentCategory`, let's refer to the correct entities as the function contains business logic

SamuelAlev (approved):

PR #338, opened 2023-02-01
chore(deps): Move sinon to devDependencies

jonasbuechel:
Move Sinon to devDependencies to prevent it from being included in the consumer dev bundle.


SamuelAlev:
Fixed in https://github.com/Frontify/brand-sdk/commit/94df6e21d931f259bf6ad5988524fb4eb08a5403

jonasbuechel:
Thanks 🙌

PR #344, opened 2023-02-09
refactor(guideline hooks): clean-up

GoranRibic:
- convert documents and `documentPages` to an array of numbers

- fix failing tests

- group `useDocuments` events to a single event

Note: I'm aware of Dummy that that is not good to pass just ids, but we have to for now. We should redefine dummy data as now is not flexible enough in most cases when we needed it. We can simplify things. Also, the property name `documentPages` to `DocumentPagesIds`, as well for `documents` is something to change in the future.


packages/app-bridge/src/react/useDocuments.ts
```diff
@@ -26,56 +21,22 @@ export const useDocuments = (appBridge: AppBridgeTheme) => {
...
-        window.emitter.on('AppBridge:GuidelineLinkAction', handleLinkEventUpdates);
-        window.emitter.on('AppBridge:GuidelineDocumentMoveAction', handleDocumentMoveEvent);
+        window.emitter.on('AppBridge:GuidelineDocumentAction', refetch);
 
         return () => {
-            window.emitter.off('AppBridge:GuidelineStandardDocumentAction', handleStandardDocumentEventUpdates);
-            window.emitter.off('AppBridge:GuidelineLibraryAction', handleLibraryEventUpdates);
-            window.emitter.off('AppBridge:GuidelineLinkAction', handleLinkEventUpdates);
-            window.emitter.off('AppBridge:GuidelineDocumentMoveAction', handleDocumentMoveEvent);
+            window.emitter.off('AppBridge:GuidelineDocumentAction', refetch);
```
SamuelAlev, line 28: Why _again_ renaming the events back to just document?

GoranRibic: Yes, as a single event simplify things, and later when every entity is a truly separate entity we will create hooks per entity and events as well. Forgot about the terrific side, thanks for the reminder 👍 on it!

packages/app-bridge/src/react/useDocumentGroups.ts
```diff
@@ -51,20 +44,31 @@ export const useDocumentGroups = (appBridge: AppBridgeTheme) => {
             });
         };
 
-        window.emitter.on('AppBridge:GuidelineDocumentGroupAction', handleEventUpdates);
```
SamuelAlev, line 54: Event still used in clarify

packages/app-bridge/src/react/useDocumentGroups.ts
```diff
@@ -51,20 +44,31 @@ export const useDocumentGroups = (appBridge: AppBridgeTheme) => {
...
         return () => {
-            window.emitter.off('AppBridge:GuidelineDocumentGroupAction', handleEventUpdates);
             window.emitter.off('AppBridge:GuidelineDocumentGroupDocumentAction', handleDocumentEventUpdates);
         };
     }, []);
 
-    return { documentGroups, refetch, isLoading };
+    /**
+     * returns list of document groups,
+     *  as default sorted by sort value
```
SamuelAlev, line 56: Could you write a better description, it is quite vague to understand.

packages/app-bridge/src/react/useDocumentGroups.ts
GoranRibic: That is correct and it should be used. It is used in useDocumentGroups as well.

SamuelAlev (approved):

PR #346, opened 2023-02-10
fix(guideline): create library

GoranRibic:
- Add Project files
- Create a project when creating a new library


packages/app-bridge/src/AppBridgeTheme.ts
```diff
@@ -114,13 +120,30 @@ export class AppBridgeTheme {
...
+
         return createDocument<DocumentLibrary>({
             ...library,
             portalId: this.getPortalId(),
-            settings: { project: library.settings?.project ?? this.getProjectId() },
+            settings: { project: projectId },
         } as DocumentLibrary);
     }
 
+    async #createProjectIdForLibrary(library: DocumentLibraryCreate) {
```
leko91, line 132: Why the # before function name?

leko91 (approved):

packages/app-bridge/src/AppBridgeTheme.ts
GoranRibic: As it should be a private method in class.

SamuelAlev: Please use `private` to align with the rest of the code.

PR #351, opened 2023-02-15
Enable `userBrandChooser` on asset chooser

findmind:
This PR enabled the brand chooser in the brand SDK (so basically it enables the brand chooser in the asset chooser for custom blocks).


findmind:
Closed due to inactivity, was done during Focus Week.

PR #365, opened 2023-03-08
feat: add targets sync events

SamCreasey:


packages/app-bridge/src/react/useDocumentPageTargets.tsx
```diff
@@ -25,9 +30,31 @@ export const useDocumentPageTargets = (appBridge: AppBridgeTheme, id: number): U
...
+            window.emitter.off('AppBridge:GuidelineDocumentPageTargetsAction', handleTargetEventUpdates);
+        };
+    }, [id]);
 
-    return { documentPageTargets, updateDocumentPageTargets, isLoading };
+    return { documentPageTargets, isLoading };
 };
+
+const updateTargets = (prevState: Nullable<DocumentPageTargets>, targetIds: number[]) =>
+    prevState
```
GoranRibic, line 51: `if(!prevState) return prevState`.

packages/app-bridge/src/react/useDocumentPages.ts
```diff
@@ -24,17 +25,30 @@ export const useDocumentPages = (appBridge: AppBridgeTheme, documentId: number)
         setIsLoading(false);
     }, [appBridge, documentId]);
 
+    const refetchIfPageExistsInMap = useCallback(
```
GoranRibic, line 28: Why refetch on Targets event? Also if refetchIfPageExistsInMap is used only in `useEffect` put it there so you avoid usage of useCallback.

packages/app-bridge/src/react/useDocumentTargets.tsx
```diff
@@ -25,9 +30,30 @@ export const useDocumentTargets = (appBridge: AppBridgeTheme, id: number): UseDo
...
+        return () => {
+            window.emitter.off('AppBridge:GuidelineDocumentTargetsAction', handleTargetEventUpdates);
+        };
+    }, [id]);
+
+    return { documentTargets, isLoading };
 };
+
+const updateTargets = (prevState: Nullable<DocumentTargets>, targetIds: number[]) =>
+    prevState
```
GoranRibic, line 50: `if(!prevState) return prevState`

packages/app-bridge/src/types/Document.ts
```diff
@@ -84,7 +85,7 @@ export type DocumentApi = Simplify<
             change_comment: Nullable<string>;
             change_comment_by: Nullable<string>;
             change_title: Nullable<string>;
-            targets: any;
+            targets: Nullable<SingleTargetApi[]>;
```
GoranRibic, line 88: This is correct, but also you need to do remapping for the `Document` type to `targets: number[]`. As we want to store only IDs.

packages/app-bridge/src/types/Emitter.ts
```diff
@@ -113,6 +113,18 @@ export type Emitter = MittEmitter<{
               action: 'delete';
           };
 
+    'AppBridge:GuidelineDocumentPageTargetsAction': {
+        targets: number[];
```
GoranRibic, line 117: the standard pattern is to use action & payload to stay consistent, so the signature, in the end, should be `action: '...'`, and the object that has `{targets: number[], pageIds: number[]} `

packages/app-bridge/src/react/useDocumentPages.ts
GoranRibic: Okay, you could do it this way. But it would be better for us to do a state update here of target ids without the additional call to API.

GoranRibic (changes requested):

packages/app-bridge/src/react/useDocumentPages.ts
SamCreasey: will be adjusted in later pr when backend is updated 👍

packages/app-bridge/src/react/useDocumentPageTargets.tsx
SamCreasey: fixed

packages/app-bridge/src/react/useDocumentTargets.tsx
SamCreasey: fixed

packages/app-bridge/src/types/Emitter.ts
SamCreasey: fixed

GoranRibic (approved):

PR #377, opened 2023-03-21
feat: Adjust document-page endpoint

leko91:


packages/app-bridge/src/react/useGuidelineActions.ts
```diff
@@ -503,7 +503,7 @@ export const useGuidelineActions = (appBridge: AppBridgeTheme) => {
 
     const moveDocumentPageBetweenDocuments = useCallback(
         async (id: number, sourceDocumentId: number, targetDocumentId: number) => {
```
oliverschwendener, line 505: Can we not remove the `sourceDocumentId` here as well? I don't think it's needed.

```suggestion
        async (id: number, targetDocumentId: number) => {
```

leko91: It is needed for the emitter, we want to update both target and source documents, and we don't know the source document otherwise

PR #397, opened 2023-04-12
feat: add guideline search signature

SamuelAlev:
Add the signature for guideline search (theme only for now)


packages/app-bridge/src/types/GuidelineSearchResult.ts
```diff
@@ -0,0 +1,21 @@
...
+    page_id: number;
+    page_slug: string;
+    page_title: string;
+    page_category_slug: string | null;
+    block_id: number;
+    document_id: number;
+    document_slug: string;
+    document_title: string;
+    portal_id: number;
+    portal_token: string;
```
oliverschwendener, line 18: This can also be null:

```suggestion
    portal_token: string | null;
```

oliverschwendener (approved):

PR #403, opened 2023-04-21
feat: Separate categorized and uncategorized pages

leko91:


packages/app-bridge/src/react/useDocumentPages.ts
```diff
@@ -12,17 +12,38 @@ export type DocumentPageEvent = {
...
 
-const sortDocumentPages = (a: DocumentPage, b: DocumentPage) => (a.sort && b.sort ? a.sort - b.sort : 0);
-
-export const useDocumentPages = (appBridge: AppBridgeBlock | AppBridgeTheme, documentId: number) => {
+export const useDocumentPages = (
+    appBridge: AppBridgeBlock | AppBridgeTheme,
+    documentId?: number,
+    categoryId?: number,
+    allPages?: boolean,
+) => {
```
SamuelAlev, line 20: The whole hook now feels hacky, please team up with someone else and try to get a cleaner hook interface

leko91: Yeah I know, that's why I left it in draft, this was more of a quick implementation to test things out in clarify.

packages/app-bridge/src/react/useDocumentPages.ts
```diff
@@ -12,17 +12,38 @@ export type DocumentPageEvent = {
...
-        const pages = await fetchDocumentPages(appBridge, documentId);
 
-        setPages(pages);
+        if (categoryId) {
+            const categorizedPages = await fetchCategorizedDocumentPages(appBridge, categoryId);
+
+            setCategorizedPages(categorizedPages);
+        }
+
+        if (documentId) {
```
GoranRibic, line 35: this logic is not ok, it will do redundant calls for all pages. If all pages are set, it should only call fetchDocumentPages. Also, the boolean allPages, should be passed in as the config object.

SamuelAlev:
Done on `main` via another PR

PR #410, opened 2023-05-03
Chore/remove depreacted document group property

oliverschwendener:
This PR changes the following types:

DocumentGroupApi:

- removed `documents` because it should not be used anymore
- added `number_of_documents`


oliverschwendener:
Ok, I give up. This thing is nested too deeply, and I have no idea how to do this API change in here.

PR #412, opened 2023-05-03
Refactor: Moving event emitters/observers out of hooks

utmishra:
### Purpose
- The aim is to have a  _separation of concern_ for event observers + emitters (plus their type definitions), thus moving them outside the React hooks.
- This allows usage/refactoring of these event methods without modifying the hooks
- Suggestion: The `events` can be renamed as `commands`, which is currently a requirement, to have a standard set of APIs to communicate with the main Frontify app.
- We can further group these events into the following:
  1. Asset Management
  2. Block and Page Settings
  3. Brand and Guideline Management
  4. File Management
  5. Navigation Management
  6. Printing Management
  7. Privacy Settings
  8. Template Management
  9. Editor State
  10. Bulk Download
- ClickUp Ticket: [#8677xch6z](https://app.clickup.com/t/8677xch6z)

### DIFF
--- a/packages/app-bridge/src/events/blockSettings.ts
+++ b/packages/app-bridge/src/events/blockSettings.ts
@@ -1,0 +1,20 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { BlockSettingsEvents, BlockSettingsUpdateCallback } from '../types';
+
+const APP_BRIDGE_BLOCK_SETTINGS_UPDATED_EVENT = BlockSettingsEvents.BlockSettingsUpdated;
+
+export const onBlockSettingsUpdated = (callback: BlockSettingsUpdateCallback) => {
+    window.emitter.on(APP_BRIDGE_BLOCK_SETTINGS_UPDATED_EVENT, callback);
+};
+
+export const offBlockSettingsUpdated = (callback: BlockSettingsUpdateCallback) => {
+    window.emitter.off(APP_BRIDGE_BLOCK_SETTINGS_UPDATED_EVENT, callback);
+};
+
+export const notifyBlockSettingsUpdated = (blockSettings: {
+    blockId: number;
+    blockSettings: Record<string, unknown>;
+}) => {
+    window.emitter.emit(APP_BRIDGE_BLOCK_SETTINGS_UPDATED_EVENT, blockSettings);
+};
--- a/packages/app-bridge/src/AppBridgeTheme.ts
+++ b/packages/app-bridge/src/AppBridgeTheme.ts
@@ -65,7 +65,12 @@ export interface AppBridgeTheme extends AppBridgeBase {
 
     deleteStandardDocument(id: number): Promise<void>;
 
-    moveDocument(documentId: number, position: number, newDocumentGroupId?: number): Promise<Document>;
+    moveDocument(
+        documentId: number,
+        position: number,
+        newDocumentGroupId?: number,
+        oldGroupId?: number,
+    ): Promise<Document>;
 
     createDocumentGroup(documentGroup: DocumentGroupCreate): Promise<DocumentGroup>;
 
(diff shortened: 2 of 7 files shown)


packages/app-bridge/src/events/blockSettings.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const APP_BRIDGE_BLOCK_SETTINGS_UPDATED_EVENT = BlockSettingsEvents.BlockSettingsUpdated;
+
+export const onBlockSettingsUpdated = (callback: BlockSettingsUpdateCallback) => {
+    window.emitter.on(APP_BRIDGE_BLOCK_SETTINGS_UPDATED_EVENT, callback);
+};
+
+export const offBlockSettingsUpdated = (callback: BlockSettingsUpdateCallback) => {
+    window.emitter.off(APP_BRIDGE_BLOCK_SETTINGS_UPDATED_EVENT, callback);
+};
```
triggertoo, line 13: This should be part of the app bridge implementation on clarify. Having this logic inside this package would mean that it's still part of the bundled block.

packages/app-bridge/src/events/blockSettings.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const notifyBlockSettingsUpdated = (blockSettings: {
+    blockId: number;
+    blockSettings: Record<string, unknown>;
+}) => {
+    window.emitter.emit(APP_BRIDGE_BLOCK_SETTINGS_UPDATED_EVENT, blockSettings);
```
triggertoo, line 19: see above :)

triggertoo (commented):
Since we are going to change the interface with the command pattern anyways, it's probably not necessary to move these emitters until we've finished it.

packages/app-bridge/src/events/blockSettings.ts
utmishra: We can have interfaces like this (names can be changed to `{dispatch/handle}Event`:

```
export interface BlockSettingsUpdateEventManager {
    onUpdate: (event: BlockSettingsUpdateEventCallback) => void;
    offUpdate: (event: BlockSettingsUpdateEventCallback) => void;
    notifyUpdate: (event: BlockSettingsUpdateEventCallback) => void;
}
```
and the implementation can be injected into the hook like:
```
export const useBlockSettings = <T = Record<string, unknown>>(
    appBridge: AppBridgeBlock,
    blockSettingsEvent: BlockSettingsUpdateEventManager,
): [T, (newSettings: Partial<T>) => Promise<void>] => {
    ....
    
    useEffect(() => {
        blockSettingsEvent.onUpdate(updateBlockSettingsFromEvent);
        
        return () => {
            blockSettingsEvent.offUpdate(updateBlockSettingsFromEvent);
        };
    
    const updateBlockSettings = async (blockSettingsUpdate: Partial<T>) => {
        try {
            await appBridge.updateBlockSettings(blockSettingsUpdate);
            blockSettingsEvent.notifyUpdate({
                blockId,
                blockSettings: mergeDeep(blockSettings, blockSettingsUpdate),
            });
        } catch (error) {
            console.error(error);
        }
    };
    
    return [blockSettings, updateBlockSettings];
    
}
```

The implementation in `clarify`:
```
import { BlockSettingsUpdateEventManager } from '@frontify/app-bridge';

export class BlockSettingsUpdateEvents implements BlockSettingsUpdateEventManager {
    public onUpdate(callback: (event: BlockSettingsUpdateEvents) => void): void {
        window.emitter.on('AppBridge:BlockSettingsUpdated', callback);
    }

    public offUpdate(callback: (event: BlockSettingsUpdateEvents) => void): void {
        window.emitter.off('AppBridge:BlockSettingsUpdated', callback);
    }

    public notifyUpdate(updatedBlockSettings: Record<string, unknown>): void {
        window.emitter.emit('AppBridge:BlockSettingsUpdated', updatedBlockSettings);
    }
}
```

utmishra: > Since we are going to change the interface with the command pattern anyways, it's probably not necessary to move these emitters until we've finished it.

Sorry, didn't notice this comment. Agreed. The example I shared above can be the _final_ version and iteratively I can start by building interfaces + implementation without touching the hooks (and replace only when finished)

utmishra:
Closing in favor of #442

PR #418, opened 2023-05-11
WIP | POC: Commands API: Dispatch based method calls for App Bridge events [#8677xch6z]

utmishra:
- A backward-compatible alternative to the present Commands API (unlike #412) 
- Adds a `dispatch` method, accepting all events with the expected payload and returning event update data, where necessary


utmishra:
Closing in favor of #442

PR #428, opened 2023-05-17
chore: Replacing private Documentation URLs with public ones

utmishra:
The Documentation links in `guideline-block-settings/src/index.ts` are private URLs (like `https://developer.frontify.com/document/1366/.....` ) and won't be accessible to the general public. Changed them with the public ones.


packages/guideline-blocks-settings/src/index.ts
```diff
@@ -27,6 +26,7 @@ import type {
     TextareaBlock as TextareaBlockSidebarSettings,
     ValueOrPromisedValue as ValueOrPromisedValueSidebarSettings,
 } from '@frontify/sidebar-settings';
+import type { FC } from 'react';
```
SamuelAlev, line 29: Please revert the change and set your IDE to not change these lines, it is noise and really not useful.

packages/guideline-blocks-settings/src/index.ts
```diff
@@ -75,34 +75,34 @@ export type BlockSettingsStructureExport = {
 export type BlockProps = {
     /**
      * The Frontify App Bridge provides an interface to the Frontify app internals.
-     * {@link https://developer.frontify.com/document/1366#/details-concepts/content-blocks/introducing-the-appbridge}
+     * {@link https://developer.frontify.com/d/XFPCrGNrXQQM/content-blocks#/details-concepts-1/content-blocks/introducing-the-app-bridge}
```
SamuelAlev, line 78: Could you check to fix the slug `detail-concepts-1` to be cleaner like `detail-concepts` ? 🤔

utmishra: This might change later anyway.
According to @floriangaechter:
> We're planning on moving the developer platform to a dedicated tool anyway sometime this year and that would involve a new path-scheme (especially in regards to the coming additions to the brand sdk)

PR #430, opened 2023-05-30
feat: New template block related interface methods

ruslan-sazonov:
This PR adds new template block related method definitions to the `AppBridgeBlock` interface:
- `getBlockTemplates()`
- `addTemplateIdsToBlockTemplateKey()`
- `deleteTemplateIdsFromBlockTemplateKey()`

Also, this PR adds new related data types and `useBlockTemplates` react hook.

[ClickUp issue](https://app.clickup.com/t/8677wr9e3)


packages/app-bridge/src/types/Template.ts
```diff
@@ -1,32 +1,16 @@
...
 
-export type Template = {
-    id: number;
-    title: string;
-    description: Nullable<string>;
-    previewUrl: string;
-    projectId: number;
-    height: number;
+export type TemplatePage = {
+    preview_url: string; // TODO: fix name inconsistency (Template.previewUrl)
```
SamuelAlev, line 13: Please map the `underscore_case` field to `camelCase`
For typing, you can do `TemplatePageApi` which can have underscore and then `type TemplatePage = CamelCasedPropertiesDeep<TemplatePage>`
And then you would need to actually map the data where you have the repository call (`web-app`) and use `convertObjectCase(result.data, 'camel');`

packages/app-bridge/src/types/TemplateLegacy.ts
```diff
@@ -0,0 +1,32 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export type TemplateApiLegacy = {
```
SamuelAlev, line 3: What is the reason why you keep `TemplateLegacy` there are no usage yet so it is safe to delete :)

ruslan-sazonov: Actually, this type is in use in `web-app -> application/client/BrandSdk/AppBridge/v3/AppBridgeBlock.ts`
We decided to keep an old template block for now for a compatibility reasons. You will see usages of that refactored legacy type in a following PR in a `web-app` repo.
Later on, when new template block will be finished and the old one will be successfully migrated we can do the cleanup.

SamuelAlev (approved):

PR #442, opened 2023-06-07
feat: Commands Interface [only for inter-platform communication calls]

utmishra:
- Based on the discussion with Marc, Leo and Luca, we have decided not to overcomplicate the Commands architecture and keep only inter-platform communication calls under Commands.
- **Structure**:
```
export const CommunicationInterface {
  dispatch(command: <CommandType>, payload?: [Optional]): CommandResponse // contains on(eventName, callback)}
}
```

The `CommandResponse` returns methods which can be subscribed to adding callback logic (Similar to [Shopify's action calls](https://shopify.dev/docs/apps/tools/app-bridge/actions/toast))


packages/app-bridge/src/commands/AssetViewer.ts
```diff
@@ -0,0 +1,7 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { Command, CommandResponse } from '../types';
+
+export interface AssetViewer {
+    dispatch(command: Command): CommandResponse[Command.AssetViewer];
```
triggertoo, line 6:
```suggestion
    dispatch(command: Command.AssetViewer): CommandResponse[Command.AssetViewer];
```

packages/app-bridge/src/types/Command.ts
```diff
@@ -0,0 +1,26 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { Asset } from './Asset';
+
+export enum Command {
+    AssetViewer = 'AssetViewer',
+    AssetChooser = 'assetChooser',
+}
```
triggertoo, line 8: it's probably better to avoid enums in these case, since it's dangerous to export this through types as it's not only a type (compile team) but also javascript code (runtime).

packages/app-bridge/src/AppBridgeBlock.ts
```diff
@@ -29,6 +29,10 @@ export interface AppBridgeBlock extends AppBridgeBase {
 
     addAssetIdsToBlockAssetKey(key: string, assetIds: number[]): Promise<void>;
 
+    /**
+     * @deprecated Use dispatch(`openAssetViewer`) command instead
+     * This will be removed in version 4.0.0 of @frontify/app-bridge
+     */
```
SamuelAlev, line 35: Keep it while making the implementation run `dispatch("openAssetViewer")` so no breaking changes

packages/app-bridge/src/commands/AssetChooser.ts
```diff
@@ -0,0 +1,7 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { Command, CommandResponse } from '../types';
+
+export interface AssetChooser {
```
SamuelAlev, line 5: type not interface

packages/app-bridge/src/commands/AssetViewer.ts
```diff
@@ -0,0 +1,7 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { Command, CommandResponse } from '../types';
+
+export interface AssetViewer {
```
SamuelAlev, line 5: type not interface

packages/app-bridge/src/commands/AssetViewer.ts
```diff
@@ -0,0 +1,7 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { Command, CommandResponse } from '../types';
```
SamuelAlev, line 3:
```suggestion
import type { Command, CommandResponse } from '../types';
```

packages/app-bridge/src/tests/AppBridgeBlockStub.ts
```diff
@@ -4,14 +4,14 @@ import mitt, { Emitter } from 'mitt';
...
+import { EmitterEvents } from '../types/Emitter';
+import { PrivacySettings } from '../types/PrivacySettings';
 import { AssetDummy } from './AssetDummy';
-import { UserDummy } from './UserDummy';
-import { ColorPaletteDummy } from './ColorPaletteDummy';
-import { ColorDummy } from './ColorDummy';
 import { BulkDownloadDummy } from './BulkDownloadDummy';
-import { PrivacySettings } from '../types/PrivacySettings';
+import { ColorDummy } from './ColorDummy';
+import { ColorPaletteDummy } from './ColorPaletteDummy';
```
SamuelAlev, line 13: Reduce noise, please set up your ide to prevent that or revert the changes

packages/app-bridge/src/tests/AppBridgeThemeStub.ts
```diff
@@ -3,10 +3,6 @@
 import mitt, { Emitter } from 'mitt';
 import { SinonStubbedInstance, spy, stub } from 'sinon';
 
-import { EmitterEvents } from '../types/Emitter';
-import { AppBridgeTheme } from '../AppBridgeTheme';
-import { mergeDeep } from '../utilities';
```
SamuelAlev, line 8: Reduce noise, please set up your ide to prevent that or revert the changes

packages/app-bridge/src/types/Command.ts
utmishra: Removed this and used direct values

packages/app-bridge/src/tests/AppBridgeThemeStub.ts
utmishra: Reverted

packages/app-bridge/src/commands/AssetViewer.ts
utmishra: Using as type now

packages/app-bridge/src/types/Command.ts
```diff
@@ -0,0 +1,21 @@
...
+export type CommandResponse = {
+    AssetViewer: {
+        on(event: CommandEvents.AssetViewerOpened, callback: () => void): void;
+        on(event: CommandEvents.AssetViewerClosed, callback: () => void): void;
+    };
+    AssetChooser: {
+        on(event: CommandEvents.AssetChooserOpened, callback: (selectedAssets: Asset[]) => void): void;
+        on(event: CommandEvents.AssetChooserClosed, callback: () => void): void;
+    };
+};
```
triggertoo, line 21:
```suggestion
export type CommandResponse = {
    AssetViewer: {
        on(event: 'assetViewerOpened', callback: () => void): void;
        on(event: 'assetViewerClosed', callback: () => void): void;
    };
    AssetChooser: {
        on(event: 'assetChooserOpened', callback: (selectedAssets: Asset[]) => void): void;
        on(event: 'assetChooserClosed', callback: () => void): void;
    };
};
```

triggertoo (commented):
no enum needed

triggertoo (approved):
lgtm 👍

utmishra:
**Update**
- The type definitions for the AssetChooser/AssetViewer commands needed update so they would be compatible with/similar to the existing AppBridge methods
- **Pending/Open question**: 
    - Hooks are publicly implemented in the AppBridge. How to leverage Command _interface_ to add the action wrapper inside these hooks

triggertoo:
not sure about that [last commit](https://github.com/Frontify/brand-sdk/pull/442/commits/95d88fe491a25ab0036aa687791a72fd511c2f6b), what's the idea of having an interface for `AssetViewer` instead using function overload on the `invoke` function within the app bridge?

triggertoo:
> **Update**
> 
> * The type definitions for the AssetChooser/AssetViewer commands needed update so they would be compatible with/similar to the existing AppBridge methods
> * **Pending/Open question**:
>   
>   * Hooks are publicly implemented in the AppBridge. How to leverage Command _interface_ to add the action wrapper inside these hooks
> 
> cc: @SamuelAlev @triggertoo

From my understanding, we don't change the way how hooks are being used. We just update the hooks to use the new command pattern.

utmishra:
Closed in favor of #460

PR #444, opened 2023-06-08
feat: add duration units to Unit type

lzurbriggen:
adds [time](https://developer.mozilla.org/en-US/docs/Web/CSS/time) units so `appendUnit` can be used for animation/transition durations.


SamuelAlev (approved):
Thank you for your contribution!

PR #446, opened 2023-06-12
feat: Template Input block refactoring

ruslan-sazonov:
This PR introduces a new `legacyTemplateInput` block type which is technically an old `templateInput` block type.


packages/sidebar-settings/src/blocks/legacyTemplateInput.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { BaseBlock } from './base';
+
+export type LegacyTemplateInputBlock<AppBridge> = {
+    /**
+     * The setting type.
+     */
+    type: 'legacyTemplateInput';
+
+    /**
+     * Whether multiple assets can be selected.
```
SamuelAlev, line 12:
```suggestion
     * Whether multiple templates can be selected.
```

ruslan-sazonov: Fixed typo

packages/sidebar-settings/src/blocks/legacyTemplateInput.ts
```diff
@@ -0,0 +1,15 @@
...
+export type LegacyTemplateInputBlock<AppBridge> = {
+    /**
+     * The setting type.
+     */
+    type: 'legacyTemplateInput';
+
+    /**
+     * Whether multiple templates can be selected.
+     */
+    multiSelection?: boolean;
```
SamuelAlev, line 14: `multiSelection` is also defined in `TemplateInputBlock` type but never used

ruslan-sazonov: I've removed `multiSelection` for now for both blocks. This prop came from an asset input block which I used as an example for a template block.

SamuelAlev (approved):

PR #454, opened 2023-06-26
feat: re-fetch on root items sorting

anxobotana:
This is an intermediate solution for the sorting root elements issues (documentGroups). 
There will be a follow up PR where the position will be preserved and the API call will just confirm the change.


fabriciomendonca (approved):
LGTM

PR #455, opened 2023-06-27
chore: add package.json script for frozen-lockfile installation

fabriciomendonca:
Adding the `ci` script to `package.json` so we can run `pnpm run ci` for installing dependencies without updating the lock file. This way, we avoid unintentional `pnpm-lock.yaml` file updates in feature or fix branches.

> The same script is used in the `web-app` and `Fondue`.


SamuelAlev:
It should not be necessary, lock file should be always up-to-date.

PR #457, opened 2023-06-28
feat: add link type and mark linkchooser as deprecated

celine-s:
creates a new link type which will look like this: 
image

can be used e.g. like this
  ```
{
            id: 'link',
            type: 'link',
            label: 'Link',
            clearable: true,
            required: true,
            defaultValue: { link: 'https://www.frontify.com', openInNewTab: false },
        },
```


SamuelAlev:
Will be done in https://github.com/Frontify/brand-sdk/pull/502

PR #467, opened 2023-07-12
feat: add info message in case browser window couldn't be opened automatically

floriangaechter:
This PR adds an info message in case the browser window couldn't be opened automatically during login (we've ran into issues with enterprise customers who had security settings in place that prevented opening browser windows from the cli).

Once the user opens the url and approves the oauth request, the rest of the process stays the same.

### DIFF
--- a/packages/cli/src/commands/login.ts
+++ b/packages/cli/src/commands/login.ts
@@ -121,8 +121,11 @@ export const loginUser = async (instanceUrl: string, port: number): Promise<void
 
         const loginUrl = authenticator.getLoginUrl();
 
-        Logger.info('Opening OAuth login page...');
+        Logger.info('Attempting to open OAuth login page...');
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/login.ts
```diff
@@ -121,8 +121,11 @@ export const loginUser = async (instanceUrl: string, port: number): Promise<void
 
         const loginUrl = authenticator.getLoginUrl();
 
-        Logger.info('Opening OAuth login page...');
+        Logger.info('Attempting to open OAuth login page...');
         await open(loginUrl);
+        Logger.info(
```
SamuelAlev, line 126: Can you switch the 2 lines?

SamuelAlev:
Please also add a changeset for the version increment `pnpm changeset`

PR #460, opened 2023-07-03
feat(Commands): Asset Chooser Interface with hooks wrapper

utmishra:
### Command Pattern-based event management for communicating with core Frontify Resources
- **Core changes**
    - Deprecating older action calls from the AppBridge, to be replaced by the `dispatch(actionHandler(options))` interface (currently replaces only `openAssetChooser`, `closeAssetChooser`)
    - Adds helper methods (`openAssetChooser` & `closeAssetChooser`) to return dispatch handlers based on the action, passing `{ commandName, options }` to the dispatch method
    - `subscribe` method to set callbacks for events
    - Hooks (`useAssetChooser` currently) will internally use the `dispatch` and `subscribe(eventName, callback)` pattern to communicate with the DOM
- How it works:
    - Trigger an action via `dispatch(eventHandler)` method
    - Subscribe to events using `subscribe` by passing the event name and expected callback
    - **Example**:
```
// Trigger an action
appBridge.dispatch(openAssetChooser(options));

appBridge.dispatch(closeAssetChooser());

// Subscribe to DOM events and pass callback
appBridge.subscribe('AssetChooser.AssetChosen', customCallback);

```

PR #472, opened 2023-07-14
refactor: Using `dispatch/subscribe` inside Asset Chooser hooks

SamuelAlev:
Extracted `useAssetChooser` hook from #460 so we don't have breaking changes waiting for the next deployments


ragi96 (approved):
🚀

PR #475, opened 2023-07-17
feat: add currentUser Method

julianiff:
Add Platform App Method:
- currentUser


packages/app-bridge/src/types/index.ts
```diff
@@ -35,3 +35,4 @@ export * from './Template';
 export * from './Terrific';
 export * from './Topic';
 export * from './User';
+export * from './PlatformAppPublicMethods';
```
SamuelAlev, line 38:
```suggestion
export * from './PlatformAppMethods';
```
No mention of public or private required here

PR #480, opened 2023-07-17
feat: platformAppBridge implementation

julianiff:


packages/app-bridge/src/types/PlatformAppMethods.ts
```diff
@@ -0,0 +1,16 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export type PlatformAppMethods = {
+    currentUser: [{ operation: 'currentUser' }, { id: string; name?: string | null; email: string; avatar: string }];
```
SamuelAlev, line 4:
```suggestion
    currentUser: [{ operation: 'currentUser' }, { id: string; name: string | null; email: string; avatar: string }];
```

packages/app-bridge/src/utilities/queryParams.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    const keyValuePairs = queryString.split('&');
+    for (const pair of keyValuePairs) {
+        const [key, value] = pair.split('=');
+        if (key && value) {
+            params[key] = value;
+        }
+    }
+
+    return params;
```
SamuelAlev, line 21:
```suggestion
export const getQueryParameters = (url: string): QueryParams => {
    const paramsToObject = (entries) => {
      const result = {}
      for(const [key, value] of entries) {
        result[key] = value;
      }
      return result;
    }

    try {
      const urlObject = new URL(url)
      return paramsToObject(urlObject.searchParams.entries());
    } catch {
      // TODO: throw new ValidationError('`url` is not a valid url')
    }
```

packages/app-bridge/src/utilities/queryParams.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+    it('should return empty object if url is undefined', () => {
+        expect(getQueryParameters(undefined)).toEqual({});
+    });
+
+    it('should return empty object if url is null', () => {
+        expect(getQueryParameters(null)).toEqual({});
+    });
+
+    it('should return objects of well formed query parameters', () => {
+        expect(getQueryParameters('https://www.frontify.com?foo=bar&baz')).toEqual({ foo: 'bar' });
```
SamuelAlev, line 28: Incorrect test case!
```suggestion
        expect(getQueryParameters('https://www.frontify.com?foo=bar&baz')).toEqual({ foo: 'bar', baz: '' });
```

packages/app-bridge/src/utilities/queryParams.ts
```diff
@@ -0,0 +1,22 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export type QueryParams = { [key: string]: string };
+
+export const getQueryParameters = (url?: string | null): QueryParams => {
```
SamuelAlev, line 5: Why allowing `undefined` and `null`

packages/app-bridge/src/AppBridgePlatformApp.ts
```diff
@@ -0,0 +1,43 @@
...
+
+type InitializeEvent = {
+    port: MessagePort;
+};
+
+export class AppBridgePlatformApp {
+    private messageBus: IMessageBus = new ErrorMessageBus();
+    private initialized = false;
+
+    public async initialize() {
```
SamuelAlev, line 19: Could probably be a `dispatch({ name: "initializePlatformApp" }): Promise<void>` (`name` field to be challenged by eco-system)

packages/app-bridge/src/AppBridgePlatformApp.ts
```diff
@@ -0,0 +1,43 @@
...
+import { getQueryParameters } from './utilities/queryParams';
+
+const PUBSUB_CHECKSUM = generateRandomString();
+
+type InitializeEvent = {
+    port: MessagePort;
+};
+
+export class AppBridgePlatformApp {
+    private messageBus: IMessageBus = new ErrorMessageBus();
```
SamuelAlev, line 16:
```suggestion
    private errorMessageBus: IMessageBus = new ErrorMessageBus();
```

packages/app-bridge/src/AppBridgePlatformApp.ts
```diff
@@ -0,0 +1,43 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { ErrorMessageBus, IMessageBus, MessageBus } from './utilities/MessageBus';
+import { generateRandomString, notify, subscribe } from './utilities';
+import { MethodsListKeys, PlatformAppMethods, Topic } from './types';
+import { InitializationError } from './errors/InitializationError';
+import { getQueryParameters } from './utilities/queryParams';
+
+const PUBSUB_CHECKSUM = generateRandomString();
```
SamuelAlev, line 9: This code is a side-effect and shouldn't be there.

SamuelAlev (changes requested):

packages/app-bridge/src/errors/InitializationError.ts
SamuelAlev, line 1: Add to sibling `index.ts`

packages/app-bridge/src/AppBridgePlatformApp.ts
marguslant0: I think this pattern is intentional. When any of the methods (`.api()`, `.context()`) are called without prior initialization, we fall back to the default `ErrorMessageBus`, otherwise we use `MessageBus`.

packages/app-bridge/src/AppBridgePlatformApp.ts
marguslant0: I think we used https://github.com/Frontify/brand-sdk/blob/9860a50edfeb15b1f5a412b778f61321b3430173/packages/app-bridge/src/AppBridgeCreateAsset.ts#L7 as an inspiration, as the `.subscribe()` method expects a token. I would be happy to improve it if there's a better way; what do you propose?

packages/app-bridge/src/AppBridgePlatformApp.ts
marguslant0: Do you mean the entire `initialize()` method would be replaced by this dispatch call? I might want to hop on a call to dig deeper on how to go about it, then :)

Just as an example how we currently figured the usage to be:

```
import "./App.css";
import { AppBridgePlatformApp } from "@frontify/app-bridge";
import { useEffect, useState } from "react";

function App() {
  const [content, setContent] = useState<any>();
  const [appBridge, setAppBridge] = useState<any>();

  useEffect(() => {
    (async () => {
      const appBridge = new AppBridgePlatformApp();
      await appBridge.initialize();
      setAppBridge(appBridge);
    })();
  }, []);

  return (
    <>
      <button
        onClick={async () => {
          const { name } = await appBridge.api({ operation: "currentUser" });
          setContent(name);
        }}
      >
        button
      </button>
      <p>{content}</p>
    </>
  );
}

export default App;

```

julianiff:
@SamuelAlev SonarCloud sees duplicate code from the Appbridge Interface that is duplicated with the AppBridgeBlock / Themes and fails as a quality gate.

julianiff:
@SamuelAlev if you want to give it a look again, i merged the main now into this branch and adjusted with the dispatch.

SamuelAlev:
Please reopen your PR, #528 somewhat corrupted the `main` branch, solving it closed the PR.

PR #483, opened 2023-07-19
feat: adding mandatory options to dispatch handler

utmishra:
- Fixes failing typecheck on `main` due to [this commit](https://github.com/Frontify/brand-sdk/commit/f2936d4274698edf9e694a2812bac61c3ece0e8f)


silviojaeger (approved):
Typecheck runs without an error now👍 As discussed, passing `options: undefined` in the helper function feels like the cleanest approach to me👍

PR #484, opened 2023-07-19
fix: type issue

SamuelAlev:
Also add a typecheck step


utmishra:
Looks good 👍🏽 
- Tested updated types with `web-app` implementation ✅ 
- Verified `typecheck` and `lint` checks (not the actions)

PR #501, opened 2023-08-07
refactor: new App Bridge interface

SamuelAlev:


SamuelAlev:
Pair programming with Henrique

PR #504, opened 2023-08-09
refactor: use Fondue serializer instead of internal version

ragi96:
If we go with https://github.com/Frontify/web-app/pull/945 we could remove the serializer here and all the hacky classes


SamuelAlev:
Please add a changeset

.changeset/cyan-ghosts-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/guideline-blocks-settings": patch
+---
+
+feat(RTE): use fondue serializer
```
SamuelAlev, line 5:
```suggestion
refactor: use Fondue serializer instead of internal version
```

SamuelAlev (approved):

SamuelAlev (changes requested):
On hold as it actually seems to break blocks

ragi96:
> On hold as it actually seems to break blocks

Not related to the changes in this PR, as it breaks (just linked) within the linkPlugin and buttonPlugin. 

A workaround is to use it within blocks as workspace package then it doesn't break. Also with version 0.28.4 it doesn't break, just linked is an issue.

fulopdaniel (approved):

PR #528, opened 2023-08-31
fix: blockItemWrapper typing issue

ragi96:
resolves that typecheck issue: 
Screenshot 2023-08-31 at 16 43 17

The problem was that with PropsWithChildren the children could be `undefined`

### DIFF
--- a/packages/guideline-blocks-settings/src/components/BlockItemWrapper/types.ts
+++ b/packages/guideline-blocks-settings/src/components/BlockItemWrapper/types.ts
@@ -1,8 +1,10 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import { MenuItemStyle } from '@frontify/fondue';
+import { ReactElement } from 'react';
 
 export type BlockItemWrapperProps = {
+    children: ReactElement;
     shouldHideWrapper?: boolean;
     shouldHideComponent?: boolean;
     toolbarItems: (ToolbarItem | undefined)[];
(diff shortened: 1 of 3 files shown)


packages/guideline-blocks-settings/src/components/BlockItemWrapper/types.ts
```diff
@@ -1,8 +1,10 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import { MenuItemStyle } from '@frontify/fondue';
+import { ReactElement } from 'react';
 
 export type BlockItemWrapperProps = {
+    children: ReactElement;
```
SamuelAlev, line 7:
```suggestion
    children: ReactNode;
```

To allow `string` otherwise feel free to keep it as is

SamuelAlev (approved):
Small comment

packages/guideline-blocks-settings/src/components/BlockItemWrapper/types.ts
```diff
@@ -1,8 +1,10 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import { MenuItemStyle } from '@frontify/fondue';
+import { ReactElement } from 'react';
```
SamuelAlev:
```suggestion
import type { ReactNode } from 'react';
```
😉

ragi96: Stupidness hit hard 🤣

PR #532, opened 2023-09-01
feat: add method on app bridge for block bulk download

SamuelAlev:
Meant to work together with Frontify/web-app#895

[Clickup](https://app.clickup.com/t/861n31u4y)


packages/app-bridge/src/AppBridgeBlock.ts
```diff
@@ -169,6 +171,10 @@ export interface AppBridgeBlock<
 
     getCurrentLoggedUser(): Promise<User>;
 
+    /**
+     * @deprecated This will be removed in version 4.0.0 of @frontify/app-bridge
+     * Use appBridge.api({ name: 'getAssetBulkDownloadToken', payload: { settingIds, documentBlockId: appBridge.getBlockId() },}); instead
+     */
```
ragi96, line 177: Not sure if needed as the hook and this function was never announced

SamuelAlev: Either you deprecate, either you delete it fully, up to you guys ;)

ragi96: We will decide within the team next week, thanks

ragi96: the old implementation is deleted

packages/app-bridge/src/registries/api/ApiMethodRegistry.ts
```diff
@@ -1,10 +1,15 @@
...
 import type { ApiMethodNameValidator } from '../../AppBridge';
+import type { GetAssetBulkDownloadTokenPayload, GetAssetBulkDownloadTokenResponse } from './GetAssetBulkDownloadToken';
 import type { GetCurrentUserPayload, GetCurrentUserResponse } from './GetCurrentUser';
 import type { GetAssetResourceInfoPayload, GetAssetResourceInfoResponse } from './GetAssetResourceInfo';
 
 export type ApiMethodRegistry = ApiMethodNameValidator<{
+    getAssetBulkDownloadToken: {
+        payload: GetAssetBulkDownloadTokenPayload;
+        response: GetAssetBulkDownloadTokenResponse;
+    };
```
julianiff, line 12: Now we share the apiMethods for blocks and platformApps. What do you think, maybe we should separate these, otherwise we have method available in the block that are not implemented.

ragi96: We just share the registry. We guard around the AppBridge with the `PlatformAppApiMethod` and `BlockApiMethod`. So I don't think we need to split it. 
Do you see something I don't see?

julianiff: ah then perfect 👍

julianiff (approved):

hochreutenerl (approved):
Nice, LGTM 👍

mike85 (approved):
LGTM!

PR #534, opened 2023-09-01
fix: added missing properties when creating document

SamuelAlev:
This PR adds missing optional properties when creating documents.


SamuelAlev:
Seems like Oli doesn't need it anymore

PR #535, opened 2023-09-01
feat: platformAppBridge implementation

SamuelAlev:


packages/app-bridge/src/AppBridgePlatformApp.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+describe('AppBridgePlatformApp', () => {
+    beforeAll(() => {
+        vi.mock('./utilities/subscribe', () => ({
+            subscribe: vi.fn().mockResolvedValue({ test: 'passed' }),
+        }));
+
+        vi.mock('./utilities/notify', () => ({
+            notify: vi.fn(),
+        }));
```
SamuelAlev, line 18: `vi.mock` doesn't need to be in a `beforeAll` ([docs](https://vitest.dev/api/vi.html#vi-mock))

packages/app-bridge/src/AppBridgePlatformApp.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    it('should be instantiable', () => {
+        expect(new AppBridgePlatformApp()).toBeInstanceOf(AppBridgePlatformApp);
+    });
+
+    it('should not initialize without a token', async () => {
+        window.location.search = '?token=';
+
+        const platformApp = new AppBridgePlatformApp();
+
+        await expect(() => platformApp.dispatch({ name: 'openConnection' })).rejects.toThrow(new InitializationError());
```
SamuelAlev, line 34:
```suggestion
        await expect(() => platformApp.dispatch(openConnection())).rejects.toThrow(new InitializationError());
```

packages/app-bridge/src/AppBridgePlatformApp.ts
```diff
@@ -16,26 +16,86 @@ import type {
...
+        };
+    }
+
+    state(): StateReturn<PlatformAppState, void>;
+    state(): unknown {
+        return undefined;
+    }
+
+    subscribe(): EventUnsubscribeFunction {
+        return () => void 0;
```
SamuelAlev, line 94: use the subscribe to know when `openConnection` is done, so we can remove `async` from `dispatch`
```ts
appBridge.dispatch(openConnection())
appBridge.subscribe('connectionOpened', () => { [...] })
```

packages/app-bridge/src/types/PlatformAppContext.ts
```diff
@@ -0,0 +1,15 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
```
SamuelAlev, line 1: What is this file?
Data like context/state must be in AppBridgePlatformApp

packages/app-bridge/src/AppBridgePlatformApp.spec.ts
julianiff: looks nice and i like the function that we can pass, but until now i didnt see this pattern (at least not with dispatch) where do we want to put these dispatch command methods?

SamuelAlev: `packages/app-bridge/src/registries/commands`

packages/app-bridge/src/AppBridgePlatformApp.ts
julianiff: Also why do we want to have openConnection() as a method? As its just an object that is returned we can also have it as a constant.

julianiff: Regarding the subscribe() pattern. As it is also not async, how do we wait in the client on this event then? 
It might be the easiest and also more understandable for the enduser, to add an initialization method that can be called that they can awaited on. It seems to me that enduser otherwise needs to know a lot to just get the appBridge started.

julianiff: Or a dispatch. Otherwise it just needs already a utility method/sugar to make it easily usable by the enduser. Lets find an easy solution here, right?

julianiff: Worked 👍

packages/app-bridge/src/types/PlatformAppContext.ts
julianiff: files moved

packages/app-bridge/src/AppBridgePlatformApp.ts
```diff
@@ -15,27 +15,141 @@ import type {
...
+            } else {
+                throw new InitializationError();
+            }
+        }
+    }
+
+    context(): ContextReturn<PlatformAppContext, void>;
+    context(): unknown {
+        return {};
+    }
```
julianiff, line 116: I will then tackle context and state in another PR

packages/app-bridge/src/AppBridgePlatformApp.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+        const platformApp = new AppBridgePlatformApp();
+        await platformApp.dispatch(openConnection());
+        expect(notify).toHaveBeenCalledTimes(1);
+    });
+
+    it('should yield true for Context.connected after dispatch', async () => {
+        window.location.search = `?token=${TOKEN}`;
+        const platformApp = new AppBridgePlatformApp();
+        platformApp.subscribe('Context.connected', () => {
+            expect(true).toBe(true);
```
SamuelAlev, line 49:
```suggestion
        platformApp.subscribe('Context.connected', () => {
            expect(connected).toBe(true);
```

packages/app-bridge/src/AppBridgePlatformApp.ts
```diff
@@ -49,11 +166,15 @@ export interface AppBridgePlatformApp<
     ): Promise<void>;
 
     state(): StateReturn<State, void>;
+
```
SamuelAlev, line 169: Spaces can be removed

packages/app-bridge/src/registries/commands/OpenConnection.ts
SamuelAlev, line 1: Filename `Connection.ts` instead? (so we could `close` maybe)

packages/app-bridge/src/AppBridgePlatformApp.ts
```diff
@@ -15,27 +15,144 @@ import type {
...
+    token: string;
     marketplaceServiceAppId: string;
+    connected: boolean;
+};
+
+export type PlatformAppContext = AppBaseProps & {
+    type: 'ASSET_ACTION';
+    assetId: string;
+    parentId: string;
+    directory: string;
```
SamuelAlev, line 54: What is `directory`?

julianiff: Its the current directory where the Asset_action Platform App is triggered from.

packages/app-bridge/src/registries/commands/Connection.ts
```diff
@@ -0,0 +1,3 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export const connection = (): { name: 'openConnection' } => ({ name: 'openConnection' });
```
SamuelAlev, line 3:
```suggestion
export const openConnection = (): { name: 'openConnection' } => ({ name: 'openConnection' });
```
Filename can be `Connection.ts` but the method should stay descriptive of the dispatch

julianiff: ah no i acutally didnt wanted to chagne the method, but idea magic just changed it xD

julianiff (approved):

packages/app-bridge/src/AppBridgePlatformApp.ts
SamuelAlev: It will grow out of hand with the more `type` you add, be careful

julianiff: Absolutly agree!

PR #538, opened 2023-09-04
feat: Error handling improvements for useBlockTemplates react hook

ruslan-sazonov:
This PR adds some minor error handling improvements to the `useBlockTemplates` react hook.

### DIFF
--- a/packages/app-bridge/src/react/useBlockTemplates.ts
+++ b/packages/app-bridge/src/react/useBlockTemplates.ts
@@ -22,6 +22,20 @@ export const useBlockTemplates = (appBridge: AppBridgeBlock) => {
         }
     };
 
+    const handleErrorMessage = (error: unknown) => {
+        let errorMessage;
+
+        if (error instanceof Error) {
+            errorMessage = error.message;
+        } else if (typeof error === 'string') {
+            errorMessage = error;
+        } else {
+            errorMessage = String(error);
+        }
+
+        setErrorMessage(errorMessage);
+    };
+
     // Fetch the block templates on mount.
     // And add listener for block template updates.
     useEffect(() => {
@@ -37,12 +51,13 @@ export const useBlockTemplates = (appBridge: AppBridgeBlock) => {
                         }
                     })
                     .catch((error) => {
-                        setErrorMessage(error as string);
+                        handleErrorMessage(error);
                     });
             };
 
-            mountingFetch();
-            window.emitter.on('AppBridge:BlockTemplatesUpdated', updateBlockTemplatesFromEvent);
+            mountingFetch().then(() =>
+                window.emitter.on('AppBridge:BlockTemplatesUpdated', updateBlockTemplatesFromEvent),
+            );
         }
 
         return () => {
@@ -57,7 +72,7 @@ export const useBlockTemplates = (appBridge: AppBridgeBlock) => {
         try {
             fetchedBlockTemplates = await appBridge.getBlockTemplates();
         } catch (error) {
-            setErrorMessage(error as string);
+            handleErrorMessage(error);
         }
 
         if (fetchedBlockTemplates) {
@@ -77,7 +92,7 @@ export const useBlockTemplates = (appBridge: AppBridgeBlock) => {
             await appBridge.deleteTemplateIdsFromBlockTemplateKey(key, oldTemplateIds);
             await appBridge.addTemplateIdsToBlockTemplateKey(key, newTemplateIds);
         } catch (error) {
-            setErrorMessage(error as string);
+            handleErrorMessage(error);
         }
 
         emitUpdatedBlockTemplates();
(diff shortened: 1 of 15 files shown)


packages/app-bridge/src/react/useBlockTemplates.ts
```diff
@@ -60,19 +77,29 @@ export const useBlockTemplates = (appBridge: AppBridgeBlock) => {
             await appBridge.deleteTemplateIdsFromBlockTemplateKey(key, oldTemplateIds);
             await appBridge.addTemplateIdsToBlockTemplateKey(key, newTemplateIds);
         } catch (error) {
-            console.error(error);
+            setErrorMessage(error as string);
```
SamuelAlev, line 80: no `as` allowed (same for the rest of the `as string`, make sure to typecheck first 
```ts
if (typeof value === 'string') {
  setErrorMessage(error)
} else {
  setErrorMessage(calculateAString)
}
```

packages/app-bridge/src/react/useBlockTemplates.ts
```diff
@@ -28,10 +29,16 @@ export const useBlockTemplates = (appBridge: AppBridgeBlock) => {
 
         if (blockId) {
             const mountingFetch = async () => {
-                const allBlockTemplates = await appBridge.getBlockTemplates();
-                if (componentMounted) {
-                    setBlockTemplates(allBlockTemplates);
-                }
+                appBridge
```
SamuelAlev, line 32: `async`/`await` please

packages/app-bridge/src/react/useBlockTemplates.ts
```diff
@@ -10,6 +10,7 @@ export const useBlockTemplates = (appBridge: AppBridgeBlock) => {
     const blockId = appBridge.getBlockId();
 
     const [blockTemplates, setBlockTemplates] = useState<Record<string, Template[]>>({});
+    const [errorMessage, setErrorMessage] = useState<string>('');
```
SamuelAlev, line 13: As you added logic, please make sure to cover it with tests

packages/app-bridge/src/react/useBlockTemplates.ts
ruslan-sazonov: Got it. Added the typechecking approach.

packages/app-bridge/src/react/useBlockTemplates.ts
ruslan-sazonov: Make sense, missed it. Thanks!

packages/app-bridge/src/react/useBlockTemplates.ts
ruslan-sazonov: Sure, just slipped out of my mind. 
Fixed the existing error-related test case and added a new one which is specific to error handling.

SamuelAlev (approved):
Looking good! 🚀

packages/app-bridge/src/react/useBlockTemplates.ts
```diff
@@ -21,21 +22,42 @@ export const useBlockTemplates = (appBridge: AppBridgeBlock) => {
...
+                        }
+                    })
+                    .catch((error) => {
+                        handleErrorMessage(error);
+                    });
             };
 
-            mountingFetch();
-            window.emitter.on('AppBridge:BlockTemplatesUpdated', updateBlockTemplatesFromEvent);
+            mountingFetch().then(() =>
```
SamuelAlev, line 58: `async`/`await` please 😬

SamuelAlev (changes requested):

packages/app-bridge/src/react/useBlockTemplates.ts
ruslan-sazonov: Sorry, I'm not getting it. What's wrong with `.then()`?

SamuelAlev: Our coding style is to use `async`/`await`

ruslan-sazonov: Ah, got it. Sorry about not following the coding style 🙈

packages/app-bridge/src/react/useBlockTemplates.ts
SamuelAlev, line 13: Suggestion:
- Rename to just `error`
- Use `null` when there are no error

```suggestion
    const [error, setError] = useState<string | null>(null);
```

SamuelAlev (approved):
Left a small comment

packages/app-bridge/src/react/useBlockTemplates.ts
ruslan-sazonov: Got it. Refactored.
I also pushed non-error related thing. 
Just found that I've missed to add the 'openNewPublication' command name to the `BlockCommand` type in `AppBridgeBlock.ts`. 
Though it will be an overkill to do the separate PR for a one liner fix.

SamuelAlev (approved):

PR #539, opened 2023-09-04
fix(RTE): link edit looks for button instead of a tags

fulopdaniel:


SamuelAlev (approved):

packages/guideline-blocks-settings/src/components/Link/utils/getUrl.ts
```diff
@@ -24,3 +24,7 @@ export const getUrl = (editor: PlateEditor) => {
 export const getUrlFromLinkOrLegacyLink = (link: TLinkElement): string => {
     return link.url || link.chosenLink?.searchResult?.link || '';
 };
+
+export const getLinkUrlFromEditor = (editor: PlateEditor) => {
```
SamuelAlev, line 28:
```suggestion
export const getLinkFromEditor = (editor: PlateEditor) => {
```

PR #554, opened 2023-09-07
feat: add items in context

SamuelAlev:
- Added `isEditing`, `isAuthenticated`, `currentLanguage`, `defaultLanguage` and `projectId` to the context
- Added helpers to build links


packages/app-bridge/src/helpers/link.ts
```diff
@@ -0,0 +1,56 @@
...
+    documentCategory?: DocumentCategory;
+    language: string;
+    portalToken: Nullable<string>;
+}) => {
+    const categoryUrlPart = documentCategory ? `${documentCategory.slug}/` : '';
+    if (portalToken) {
+        return `/d/${portalToken}/${language}/${document.slug}#/${categoryUrlPart}${documentPage.slug}`;
+    }
+
+    return `/document/${document.id}/${language}#/${categoryUrlPart}${documentPage.slug}`;
```
hochreutenerl, line 55: Please use `getUrlFragment` for the hash and following parts. Same for line 52 above

SamuelAlev: We are in Brand SDK repository, not Web App. Did you change some routing logic recently?

hochreutenerl: Yes. The `categoryUrlPart` should be `-/` if there is no `documentCategory`
From my current perspective we should avoid generating these urls externally.

packages/app-bridge/src/AppBridgeBlock.ts
```diff
@@ -204,22 +204,10 @@ export interface AppBridgeBlock<
      */
     closeAssetChooser(): void;
 
-    /**
```
SamuelAlev, line 207: To be done later

packages/app-bridge/src/AppBridgeTheme.ts
```diff
@@ -252,7 +259,7 @@ export interface AppBridgeTheme<
 
     updateBrandportalLink(brandportalLink: Partial<BrandportalLink>): Promise<Partial<BrandportalLink> | null>;
 
-    getCoverPage(): Promise<CoverPage>;
+    getCoverPage(language?: string): Promise<CoverPage>;
```
mike85, line 262: Does the caller of this function need to have control over the `language` parameter?

packages/app-bridge/src/react/useCoverPage.ts
```diff
@@ -26,7 +26,7 @@ export const useCoverPage = (
 
     const fetchCoverPage = useCallback(async () => {
         setIsLoading(true);
-        setCoverPage(await appBridge.getCoverPage());
+        setCoverPage(await appBridge.getCoverPage(appBridge.context('currentLanguage').get()));
```
mike85, line 29: Couldn’t we just call `getCoverPage()` and use the updated context value from within the appBridge function?

SamuelAlev: By default it uses the context in the implementation, so it is not necessary here.

Screenshot 2023-11-29 at 14 36 41

Will remove it 👍

packages/app-bridge/src/AppBridgeTheme.ts
SamuelAlev: Removed

PR #559, opened 2023-09-11
fix: buttonPlugin margins

ragi96:
add margin-top/bottom to buttons so they behave the same as in the old RTE


fulopdaniel (approved):

mike85 (approved):
Looks good to me!

PR #563, opened 2023-09-13
fix(RTE): edit modal styles can be invisible

fulopdaniel:
The edit modals were using the `BlockStyles.p` styles, which can be white color. My opinion is that the RTE edit modal styles should be consistent, regardless of the design settings, as that is only used for editing. Open for debating.

Screenshot 2023-09-13 at 08 21 14
Screenshot 2023-09-13 at 08 19 28

[23021/TASK-8529](https://app.clickup.com/t/2523021/TASK-8529)


packages/guideline-blocks-settings/src/components/RichTextEditor/plugins/ButtonPlugin/components/FloatingButton/EditButtonModal/EditModal.tsx
```diff
@@ -2,17 +2,17 @@
```
ragi96, line 2: add a changeset

fulopdaniel: added

anxobotana (approved):
LGTM

PR #575, opened 2023-09-18
feat: add init hook to react

julianiff:


packages/app-bridge/src/react/usePlatformAppBridge.ts
```diff
@@ -0,0 +1,20 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { AppBridgePlatformApp } from '../AppBridgePlatformApp';
+import { useEffect, useState } from 'react';
+
+export const usePlatformAppBridge = () => {
+    const [platformAppBridge] = useState<AppBridgePlatformApp>(new AppBridgePlatformApp());
```
marguslant0, line 7: Does it need to be done via `useState`? Would simply `const platformAppBridge = new AppBridgePlatformApp();` also work?

julianiff: Good remark. It just acts like a "singelton" pattern. Due to connection updates that are happening it could create 2 platformAppbridge objects. To prevent that in React i used the useState to not create multiple objects when a rerender is happening

marguslant0 (approved):
💪

mike85 (approved):
LGTM!

julianiff:
cc @getflourish react utility method to initiate a platformAppBridge

PR #576, opened 2023-09-20
feat: add localContext set in init and add get / subscribe methods.

julianiff:
Web-App Part https://github.com/Frontify/web-app/pull/1241

### DIFF
--- a/packages/app-bridge/src/AppBridgePlatformApp.spec.ts
+++ b/packages/app-bridge/src/AppBridgePlatformApp.spec.ts
@@ -10,7 +10,7 @@ const TOKEN = 'AjY34F87Dsat^J';
 
 describe('AppBridgePlatformApp', () => {
     vi.mock('./utilities/subscribe', () => ({
-        subscribe: vi.fn().mockResolvedValue({ test: 'passed' }),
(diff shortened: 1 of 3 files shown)


packages/app-bridge/src/AppBridgePlatformApp.spec.ts
```diff
@@ -64,4 +64,24 @@ describe('AppBridgePlatformApp', () => {
...
+            expect({ parentId: 'parentId-test' }).toStrictEqual(context);
+        });
+        platformApp.context('parentId').subscribe((parentId) => {
+            expect('parentId-test').toStrictEqual(parentId);
+        });
+        platformApp.subscribe('Context.connected', () => {
+            const context = platformApp.context().get();
+            const parentId = platformApp.context('parentId').get();
+
+            expect(connected).toBe(true);
```
marguslant0, line 81: What's the idea here? 🤔  On line 68 we set the value to `true` ourselves..

julianiff: ups :D that should have been the callback

marguslant0 (approved):
image

PR #585, opened 2023-09-26
fix: update react hook to return undefined and appBridge

julianiff:
…ws easier handling in the app. 

@getflourish 

Now we can do 
```
const appBridge = usePlatformAppBridge();
[…]
const context = appBridge?.context().get();
```

### DIFF
--- a/packages/app-bridge/src/react/usePlatformAppBridge.spec.ts
+++ b/packages/app-bridge/src/react/usePlatformAppBridge.spec.ts
@@ -19,18 +19,14 @@ describe('usePlatformAppBridge', () => {
     afterEach(() => {
         vi.clearAllMocks();
     });
-    it('should return platformAppBridge', async () => {
+    it('should return undefined platformApp if not initiated', async () => {
         const { result } = renderHook(() => usePlatformAppBridge());
-        expect(result.current.platformAppBridge).toBeDefined();
+        expect(result.current).toBeUndefined();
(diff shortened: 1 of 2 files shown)


packages/app-bridge/src/react/usePlatformAppBridge.spec.ts
```diff
@@ -19,18 +19,14 @@ describe('usePlatformAppBridge', () => {
...
-    it('should return initiated false', async () => {
-        const { result } = renderHook(() => usePlatformAppBridge());
-        expect(result.current.connected).toBe(false);
-    });
-    it('should return initiated true after waiting', async () => {
+    it('should return platformApp after initiation and waiting', async () => {
         const { result } = renderHook(() => usePlatformAppBridge());
         await waitFor(() => {
-            expect(result.current.connected).toBe(true);
+            expect(result.current).toBeTypeOf('object');
```
marguslant0, line 29: Would it make sense to also add the earlier
```
expect(result.current.platformAppBridge).toBeDefined();
```
to make it more specific?

julianiff:
```suggestion
            expect(result.current).toBeDefined();
```

mike85 (approved):
LGTM!

marguslant0 (approved):
image

PR #590, opened 2023-09-27
feat: add messagequeue to simulate sync message sending

julianiff:
Added messagequeue instead of messsageChannel to have a sync behaviour


packages/app-bridge/src/utilities/MessageQueue.spec.ts
```diff
@@ -1,14 +1,14 @@
...
-import { MessageBus } from './MessageBus';
+import { MessageQueue } from './MessageQueue';
 
 describe('MessageBus', () => {
     it('should be instantiable', () => {
         const { port1 } = new MessageChannel();
 
-        const messageBus = new MessageBus(port1);
-        expect(messageBus).toBeInstanceOf(MessageBus);
+        const messageBus = new MessageQueue(port1);
```
marguslant0, line 10: Might as well rename the local variables here also to `messageQueue` for consistency

julianiff:
closed in favor of tokens

PR #593, opened 2023-09-28
chore: make the ecosystem team the code owner of the types

hochreutenerl:
Currently G2 is still owner of the types in the brand sdk. From my understanding this should now also be part of ecosystem.

Also I saw that for Web-App and App-Server the Codeowners file is owned by the architecture team. I assume, this should be the same here, right?


mike85 (approved):
LGTM

agjacome (approved):

henser (approved):

PR #597, opened 2023-10-05
feat: extended guideline search result type

bojangles-m:


SamuelAlev (approved):

bojangles-m:
@Frontify/ecosystem hi, could please one review this PR?

henser (commented):
I'm too foreign to this at the moment but I have some questions about the typing here ... and maybe on the existing/previous implementation already.
It seems to me this `GuidelineSearchResultApi` type should provide some sort of abstraction/interface where only the relevant properties for each of it's implementations are applied 🤔  Is it the case that these new `section_*` properties are only relevant for `type` `SECTION` results?

SamuelAlev:
@henser I will merge the PR like it is if you are fine, we anyway need to migrate the method from `appBridge.searchInGuideline` to something along `appBridge.api({ name: "search", payload: [...] })`

henser (approved):
Sure @SamuelAlev👌 As I said I'm too foreign to this change to be blocking development here 👍 Approved ✅

PR #598, opened 2023-10-06
feat(Dropdown): extend block with position

fulopdaniel:


packages/sidebar-settings/src/blocks/dropdown.ts
```diff
@@ -24,6 +24,11 @@ export type DropdownBlock<AppBridge> = {
      */
     size?: 'small' | 'large' | DropdownSize;
 
+    /**
+     * Whether the dropdown should open to top or to bottom.
+     */
+    position?: 'top' | 'bottom';
```
SamuelAlev, line 30: The devs shouldn't be able to control that, it should be automatic in the sidebar

fulopdaniel: image

For me it always drops down even if there is limited space. The height of the dropdown area is dynamic, but the position seems to be hardcoded to bottom.

fulopdaniel: We can go for an approach where we tweak the component in Fondue to automatically open to a position where there is more space.

SamuelAlev: Should be fixed within the sidebar instead of the settings (maybe the Fondue Dropdown already support the `flip` option?)

fulopdaniel: I will close this PR and fix it in Fondue.

PR #599, opened 2023-10-09
feat: add flag to run serve as a platformApp

julianiff:


julianiff:
@marguslant0 and @SamuelAlev i added the optional --appType flag and the read on the manifest to decide on the devServer that will be run.

packages/cli/src/index.ts
SamuelAlev, line 1: Could you add a changeset?

packages/cli/src/index.ts
```diff
@@ -78,8 +80,16 @@ cli.command('serve', 'serve the app locally')
     .option('--allowExternal, --allow-external', '[boolean] allow external IPs to access the server', {
         default: false,
     })
+    .option('--appType [appType], --app-type', '[string] specify app type. Overrides manifest values')
```
marguslant0, line 83: Is there a reason why `[appType]` is in brackets? 🤔 

Should it perhaps be:
`.option('--appType, --app-type, <appType>', ...` instead? (taking inspiration from the `entryPath`)

packages/cli/src/index.ts
```diff
@@ -78,8 +80,16 @@ cli.command('serve', 'serve the app locally')
...
     .action(async (options) => {
-        await createDevelopmentServer(options.entryPath, options.port, options.allowExternal);
+        const manifest = reactiveJson<AppManifest>(join(process.cwd(), 'manifest.json'));
+        const appType = options.appType ?? manifest.appType;
```
SamuelAlev, line 86:
```suggestion
        const appType = options.appType || manifest.appType;
```
`'' | 'theme' === ''`

packages/cli/src/index.ts
SamuelAlev: @julianiff Still open?

julianiff: ahh brackets means its optional. sorry we cleared it in a 1on1. 
From [cac npm](https://www.npmjs.com/package/cac#brackets): 
When using brackets in command name, angled brackets indicate required command arguments, while square bracket indicate optional arguments.

PR #604, opened 2023-10-17
fix: deprecated dropdown types

floriangaechter:
This change fixes the deprecated dropdown types in the block templates


packages/cli/templates/content-block-css-modules/src/settings.ts
SamuelAlev, line 1: Can you add a changeset? `pnpm changeset`

SamuelAlev (approved):

PR #610, opened 2023-10-23
fix(cli): never ignore manifest.json

ragi96:
Just had a case in which an agency had the manifest.json in the `.gitignore`, so the file wasn't deployed and the Block data wasn't updated. So this PR just filters the `manifest.json` out of the `gitignoreEntries`, so it's never a problem. 

More Context:
- They deploy the Block on multiple instances and have an AppId for each, so they have multiple `manifest.json`'s 
- So they have a script deploy:instanceA which then renames the instancea.manifest.json to manifest.json and deploys

### DIFF
--- a/.changeset/curly-tips-turn.md
+++ b/.changeset/curly-tips-turn.md
@@ -1,0 +1,5 @@
+---
+"@frontify/frontify-cli": patch
+---
+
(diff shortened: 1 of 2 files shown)


mike85 (approved):
LGTM!

PR #611, opened 2023-10-23
feat: Add language to getAssetBulkDownloadToken API (and pass it via hook)

utmishra:
- Adds a `language` to `getAssetBulkDownloadToken` AppBridgeBlock `api`
- Passes the new language parameter via `useAssetBulkDownload` hook

### DIFF
--- a/packages/app-bridge/src/react/useAssetBulkDownload.spec.tsx
+++ b/packages/app-bridge/src/react/useAssetBulkDownload.spec.tsx
@@ -7,6 +7,7 @@ import { getAppBridgeBlockStub } from '../tests';
 import { AssetBulkDownloadState, useAssetBulkDownload } from './';
 
 const appBridgeError = new Error('Something went wrong');
+const DUMMY_LANGUAGE = 'de';
 
 describe('useAssetBulkDownload', () => {
     beforeEach(() => {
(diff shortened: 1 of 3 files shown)


julianiff (approved):

mike85 (approved):
LGTM

packages/app-bridge/src/react/useAssetBulkDownload.spec.tsx
```diff
@@ -7,6 +7,7 @@ import { getAppBridgeBlockStub } from '../tests';
 import { AssetBulkDownloadState, useAssetBulkDownload } from './';
 
 const appBridgeError = new Error('Something went wrong');
+const DUMMY_LANGUAGE = 'de';
```
hochreutenerl, line 10: This const seems unused? On purpose?

packages/app-bridge/src/react/useAssetBulkDownload.ts
```diff
@@ -22,7 +22,11 @@ export const useAssetBulkDownload = (appBridge: AppBridgeBlock) => {
 
             const { assetBulkDownloadToken } = await appBridge.api({
                 name: 'getAssetBulkDownloadToken',
-                payload: { settingIds, documentBlockId: appBridge.getBlockId() },
+                payload: {
+                    settingIds,
+                    documentBlockId: appBridge.getBlockId(),
+                    language: appBridge.getTranslationLanguage(),
```
hochreutenerl, line 28: Currently, I think we don't even need this PR. Did you check the possibility to just add the language parameter on the web-app side of the app bridge, similar like we did for other similar cases?
image

Probably we need a refactoring to use the `getTranslationLanguage` within the new API structure.

utmishra:
Closing this as we fetch language [directly from AppBridge now](https://github.com/Frontify/web-app/pull/1701)

PR #613, opened 2023-10-24
chore: add hint to readme for local environment

hochreutenerl:
Title says it all 😄

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -60,6 +60,8 @@ The Frontify URL represent your instance URL.
 
 Example: `weare.frontify.com` or `https://weare.frontify.com`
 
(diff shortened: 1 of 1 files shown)


packages/cli/README.md
```diff
@@ -60,6 +60,8 @@ The Frontify URL represent your instance URL.
 
 Example: `weare.frontify.com` or `https://weare.frontify.com`
 
+If you are using a local development environment, you might need to put `NODE_TLS_REJECT_UNAUTHORIZED=0` before the command
```
SamuelAlev, line 63:
```suggestion
If you are using a domain with self-signed certificate, you will need to put `NODE_TLS_REJECT_UNAUTHORIZED=0` before the log in command
```

SamuelAlev (approved):

PR #616, opened 2023-10-25
feat: Add state channel and logic to update state in web-app

julianiff:


packages/app-bridge/src/AppBridgePlatformApp.spec.ts
```diff
@@ -62,10 +64,22 @@ describe('AppBridgePlatformApp', () => {
...
+    it('should return state after app is initialized', async () => {
+        window.location.search = `?token=${TOKEN}`;
+        const platformApp = new AppBridgePlatformApp();
+        platformApp.subscribe('Context.connected', (connected) => {
+            const state = platformApp.state().get();
+
+            expect(connected).toBe(true);
+            expect(state).toEqual({ settings: 'settings-test' });
+        });
+        platformApp.dispatch(openConnection());
```
marguslant0, line 82: This probably has no effect on the test, and can be removed?

marguslant0 (approved):
Looks good! 👍

packages/app-bridge/src/AppBridgePlatformApp.spec.ts
julianiff: Should throw an error when the setting object is not correct, did they not do anything for you?

marguslant0: No errors for me 🤷 The expectations have already run, so it shouldn't have any effect

julianiff: Screenshot 2023-10-27 at 10 43 55

For me it throws an assertionError, hm

marguslant0: what is the asdf-ending settings value?

For me it passes nicely when I run it on terminal 🤷 
image

julianiff: Ahhh i see, if you dont put the dispatch then the subscribe doesnt trigger. Its a pub sub pattern, and the dispatch initiates the "publishing" here. Only when the dispatch publishes something then the subscribe is triggered.

marguslant0: Ah, I see. Would it then make sense to move the dispatch before the subscribe?

julianiff: First you need to subscribe to smth, then you can dispatch a method on it.

julianiff: If you put the dispatch in front of the subscribe, then the test doesnt fail because the subscribe is not executed, i would assume, but didnt test it

packages/app-bridge/src/AppBridgePlatformApp.ts
```diff
@@ -94,35 +98,45 @@ export class AppBridgePlatformApp implements IAppBridgePlatformApp {
...
         apiHandler: ApiHandlerParameter<ApiMethodName, PlatformAppApiMethod>,
     ): ApiReturn<ApiMethodName, PlatformAppApiMethod> {
-        return this.messageBus.post({
-            method: 'api',
+        return this.apiMessageBus.post({
             parameter: apiHandler,
         }) as ApiReturn<ApiMethodName, PlatformAppApiMethod>;
     }
 
+    private initializationGuard() {
```
marguslant0, line 106: As it's a method name, let's use imperative mode, e.g. `checkInitialization()`

packages/app-bridge/src/AppBridgePlatformApp.ts
```diff
@@ -94,35 +98,45 @@ export class AppBridgePlatformApp implements IAppBridgePlatformApp {
...
+        const initialContext = getQueryParameters(window.location.href);
+        if (!initialContext.token) {
+            throw new InitializationError();
+        }
+        if (this.initialized) {
```
marguslant0, line 111: What if `this.initialized` evaluates falsy? We would still return `undefined`. Should it also throw then?

packages/app-bridge/src/AppBridgePlatformApp.ts
julianiff: Thats a very good point. 

The idea with this function is to Guard against programm execution. When not all the statements are fulfilled it should abort (silently or with an error in this case). Here we use [Guard as a Pattern](https://en.wikipedia.org/wiki/Guard_(computer_science)). let me think a moment if we can do both, be explicit about it beeing a guard and having a verb involved. If you have any idea im a tacker :D

packages/app-bridge/src/AppBridgePlatformApp.ts
julianiff: When developing an app you have the strict mode that rerenders twice the apps. If that happens the object is already initiated and doesnt need another initiation (It already has the needed objects from the web-app). It can just be ignored and nothing needs to happen.

packages/app-bridge/src/AppBridgePlatformApp.ts
marguslant0: Thanks for elaborating! 

'guard' itself is also a verb, perhaps something starting with `guardFor..`?

mike85 (approved):
The code owned by G2 team looks good to me

PR #626, opened 2023-11-06
feat: add url rule

mike85:
This PR adds the `urlRule` that validates url strings.

### DIFF
--- a/packages/sidebar-settings/src/helpers/rules/urlRule.ts
+++ b/packages/sidebar-settings/src/helpers/rules/urlRule.ts
@@ -1,0 +1,15 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { Rule } from './Rule';
+
+export const urlRule: Rule<string> = {
+    errorMessage: 'Please use a valid url',
+    validate: (value: string): boolean => {
+        const regex = new RegExp(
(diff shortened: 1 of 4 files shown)


packages/sidebar-settings/src/helpers/rules/urlRule.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const urlRule: Rule<string> = {
+    errorMessage: 'Please use a valid url',
+    validate: (value: string): boolean => {
+        const regex = new RegExp(
+            '^((http|https)://)[-a-zA-Z0-9@:%._+~#?&/=]{2,256}\\.[a-z]{2,6}\\b([-a-zA-Z0-9@:%._+~#?&/=]*)$',
+            'i',
+        );
+
+        return regex.test(value);
```
SamuelAlev, line 13:
```suggestion
        try {
          new URL(value)
          return true
        } catch {
           return false
        }
```

mike85: This was indeed my first approach but as it considers e.g. `https://index.html` as a valid url, I decided to go the regex approach to be more restrictive. What’s your opinion here?

SamuelAlev: This regex approach prevent custom protocols like `frontify://login`

mike85: That’s true. So we’ll follow the approach to allow all possible valid urls. Thanks!

SamuelAlev (approved):

PR #627, opened 2023-11-07
feat: privacy settings async

ragi96:
Changes the handling to get the privacySettings to an async approach.

Also changes the stub of the `.api` to the newer approach with `callsFake` to have more control than `withArgs` as it wasn't not possible to handle both

@SamuelAlev not sure at what time we should remove the old `getPrivacy()` from the Interface and afterwards from the implementation so, we can remove the context appearance usage.

cc: @jonasbuechel 

Meant to work together with: https://github.com/Frontify/web-app/pull/1898


julianiff (approved):

packages/app-bridge/src/react/usePrivacySettings.ts
```diff
@@ -7,7 +7,22 @@ import type { PrivacySettings } from '../types/PrivacySettings';
...
+    useEffect(() => {
+        const fetchPrivacySettings = async () => {
+            const privacySettings = await appBridge.api({
+                name: 'getPrivacySettings',
+                payload: { portalId: appBridge.context('portalId').get() },
+            });
+            setPrivacySettings(privacySettings);
+        };
+        fetchPrivacySettings();
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
SamuelAlev, line 24: Please add the dependency

packages/app-bridge/src/registries/api/GetPrivacySettings.spec.ts
```diff
@@ -0,0 +1,19 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { describe, expect, it } from 'vitest';
+import { getPrivacySettings } from './GetPrivacySettings';
+
+describe('getPrivacySettings', () => {
+    it('should create an action to get privacy settings with the correct name and payload', () => {
+        const PORTAL_ID = 123;
```
SamuelAlev, line 8: Can be moved outside the `describe`

packages/app-bridge/src/tests/AppBridgeBlockStub.ts
```diff
@@ -213,7 +219,6 @@ export const getAppBridgeBlockStub = ({
         getDocumentTargets: stub<Parameters<AppBridgeBlock['getDocumentTargets']>>().resolves(),
         getDocumentPageTargets: stub<Parameters<AppBridgeBlock['getDocumentPageTargets']>>().resolves(),
         state: stub<Parameters<AppBridgeBlock['state']>>().resolves(),
-        context: stub<Parameters<AppBridgeBlock['context']>>().resolves(),
```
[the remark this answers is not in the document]
ragi96: Isn't removed, just moved up (Line 189), as I added the stub for getting the portal via the context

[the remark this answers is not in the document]
SamuelAlev: Ok I might need to buy a pair of glasses

packages/app-bridge/src/react/usePrivacySettings.ts
```diff
@@ -7,7 +7,21 @@ import type { PrivacySettings } from '../types/PrivacySettings';
...
+
+    useEffect(() => {
+        const fetchPrivacySettings = async () => {
+            const privacySettings = await appBridge.api({
+                name: 'getPrivacySettings',
+                payload: { portalId: appBridge.context('portalId').get() },
+            });
+            setPrivacySettings(privacySettings);
+        };
+        fetchPrivacySettings();
```
fulopdaniel, line 23: Optional.
To me this is a bit more readable, as fetchPrivacySettings does not imply that it also sets the state. `fetchPrivacySettings` could also potentially be moved to a helper function.

```suggestion
        const fetchPrivacySettings = async () => {
            return await appBridge.api({
                name: 'getPrivacySettings',
                payload: { portalId: appBridge.context('portalId').get() }
            });
        };
        setPrivacySettings(await fetchPrivacySettings());
```

fulopdaniel (approved):

packages/app-bridge/src/react/usePrivacySettings.ts
ragi96: True it would be better readable, but it isn't that easy as `Effect callbacks are synchronous to prevent race conditions. Put the async function inside`, so I will leave it as it is as its easier to read than changing this

PR #630, opened 2023-11-09
feat: show more information in the error message when login fails

floriangaechter:


packages/cli/src/commands/login.ts
```diff
@@ -68,8 +68,8 @@ export class Authenticator {
                 '/api/oauth/random',
             );
             this.randomChallenge = randomCodeChallenge.data;
-        } catch {
-            throw new Error('An error occured while getting the random challenge.');
+        } catch (error) {
+            throw new Error(`An error occured while getting the random challenge: ${(error as Error).message}`);
```
SamuelAlev, line 72: `error` might not be `Error`, surround the line with `if` to make sure (and then we can remove the `as Error`)

packages/cli/src/commands/login.ts
```diff
@@ -126,8 +126,8 @@ export const loginUser = async (instanceUrl: string, port: number): Promise<void
             `If a browser window doesn't automatically open, please open the following link manually: ${loginUrl}`,
         );
         await open(loginUrl);
-    } catch {
-        Logger.error('You need to enter a valid Frontify instance URL.');
+    } catch (error) {
+        Logger.error(`You need to enter a valid Frontify instance URL: ${(error as Error).message}`);
```
SamuelAlev, line 130: Same here

packages/cli/src/commands/login.ts
```diff
@@ -69,7 +69,9 @@ export class Authenticator {
             );
             this.randomChallenge = randomCodeChallenge.data;
         } catch (error) {
-            throw new Error(`An error occured while getting the random challenge: ${(error as Error).message}`);
+            if (error instanceof Error) {
+                throw new TypeError(`An error occured while getting the random challenge: ${error.message}`);
+            }
```
SamuelAlev, line 74:
```suggestion
            if (error instanceof Error) {
                throw new TypeError(`An error occured while getting the random challenge: ${error.message}`);
            }
            throw new TypeError(`An unknown error occured while getting the random challenge.`);
```

SamuelAlev (approved):

PR #637, opened 2023-11-23
feat: add portal languages to AppBridgeTheme context

mike85:
This PR adds the portal languages to the AppBridgeTheme context.


packages/app-bridge/src/types/Brandportal.ts
```diff
@@ -0,0 +1,20 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export type BrandportalApi = {
+    i18n_settings: {
+        languages: BrandportalI18nLanguage[];
+    };
+};
+
+export type BrandportalI18nLanguage = {
+    language: string;
```
olivereisenhut, line 10: Is this an ISO language code?

packages/app-bridge/src/types/Brandportal.ts
```diff
@@ -0,0 +1,20 @@
...
+    label: string;
+    default?: boolean;
+    draft?: boolean;
```
olivereisenhut, line 13: Could those be defaulted to false instead of having them nullable?

mike85: I changed the type accordingly.

packages/app-bridge/src/types/Portal.ts
```diff
@@ -0,0 +1,8 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export type PortalLanguage = {
+    isoCode: string;
+    name: string;
+    isDefault: boolean;
+    isDraft: boolean;
```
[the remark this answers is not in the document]
mike85: In general, I find it good practice if you can already identify a boolean type via the property name. However I’ll change it to `default` and `draft` as I just realised that we don’t have any booleans with the `is` prefix in the AppBridge.

olivereisenhut (approved):

packages/app-bridge/src/types/Portal.ts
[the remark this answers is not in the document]
SamuelAlev: Then let's keep the `is` prefix 👍

packages/app-bridge/src/types/Language.ts
```diff
@@ -0,0 +1,23 @@
...
+     */
+    name: string;
+
+    /**
+     * Indicates whether the language is the default language.
+     */
+    isDefault: boolean;
+
+    /**
+     * Indicates whether the language is in draft status.
```
SamuelAlev, line 20:
```suggestion
     * Indicates whether the language is in draft state.
```

SamuelAlev (approved):

PR #642, opened 2023-12-01
chore: move into common create folder

julianiff:
- Moved create commands into own create file to have the same structure as the other commands.


packages/cli/src/commands/create.ts
```diff
@@ -1,12 +1,11 @@
...
 
+import { Logger, copyFolder, updatePackageJsonProjectName } from '../utils/index.js';
+import pc from 'picocolors';
 import { resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
-import pc from 'picocolors';
-import { Logger, copyFolder, updatePackageJsonProjectName } from '../utils/index.js';
 
 const CONTENT_BLOCK_PREFIX = 'content-block-';
-
```
SamuelAlev, line 9: Can be put back

packages/cli/src/commands/create.ts
```diff
@@ -26,3 +25,22 @@ export const createNewContentBlock = (contentBlockName: string, template = 'tail
 
     Logger.defaultInfo(`\n${Logger.spacer(11)}Happy hacking!`);
 };
+export const createNewApp = (appName: string, template, type: string): void => {
```
SamuelAlev, line 28:
```suggestion
};

export const createNewPlatformApp = (appName: string, template, type: string): void => {
```

julianiff: Ive just seen, these 2 things are so much the same, they should not coexist.. if its fine that a response in the cli says that the "Scaffolding App..." instead of "Scaffolding Block" then we could just use only one method

packages/cli/src/index.ts
```diff
@@ -238,7 +237,7 @@ for (const appType of ['block', 'theme']) {
     cli.command(
         `${appType} create [appName]`,
         `[deprecated: use 'create' instead] create a ${appType} app locally`,
-    ).action((appName: string) => createNewContentBlock(appName, 'css-modules'));
+    ).action((appName: string) => createNewApp(appName, 'css-modules', 'content-block'));
```
SamuelAlev, line 240:
```suggestion
    ).action((appName: string) => createNewApp(appName, 'css-modules', appType));
```

packages/cli/src/index.ts
```diff
@@ -227,7 +226,7 @@ cli.command('create [appName]', 'create a new marketplace app')
                 exit(0);
             }
 
-            createNewContentBlock(promptedAppName, stylingFramework);
+            createNewApp(promptedAppName, stylingFramework, 'content-block');
```
SamuelAlev, line 229: ?
```suggestion
            createNewApp(promptedAppName, stylingFramework, appType);
```

PR #654, opened 2023-12-15
feat: themes and template settings

anxobotana:


mike85 (approved):
LGTM (fixed everything during pair review)

packages/app-bridge/src/types/Emitter.ts
```diff
@@ -39,6 +40,7 @@ export type EmitterEvents = {
         portalId: number;
         themeAssets: Record<string, Asset[]>;
         prevThemeAssets: Record<string, Asset[]>;
+        template: ThemeTemplate;
```
SamuelAlev, line 43: Theme assets should be applied regardless of the template, why adding the info?

packages/app-bridge/src/types/Emitter.ts
```diff
@@ -13,12 +13,13 @@ import type { DocumentCategory } from './DocumentCategory';
...
 import type { PrivacySettings } from './PrivacySettings';
 import type { Template } from './Template';
+import type { ThemeTemplate } from './ThemeTemplate';
 
 export type EmitterAction = 'add' | 'update' | 'delete';
 
 export type EmitterEvents = {
     'AppBridge:PageTemplateSettingsUpdated': { pageTemplateSettings: Record<string, unknown> };
-    'AppBridge:ThemeSettingsUpdated': { themeSettings: Record<string, unknown> };
+    'AppBridge:ThemeSettingsUpdated': { themeSettings: Record<ThemeTemplate, Record<string, unknown>> };
```
SamuelAlev, line 22: Please explain the changes, `theme settings !== page template settings`

```
                 theme
                   |
         |----------------------|
theme settings    page template settings
                                 |
           |---------------------|------------------|
document page         cover page            library page
template settings     template settings     template settings
```

packages/app-bridge/src/tests/AppBridgeThemeStub.ts
```diff
@@ -80,7 +80,7 @@ export const getAppBridgeThemeStub = ({
     projectId = PROJECT_ID,
     pageTemplateSettings = {},
```
SamuelAlev, line 81: The settings for the different templates should be here

SamuelAlev (changes requested):

packages/app-bridge/src/types/Emitter.ts
anxobotana: `theme settings` are now per template, we have:
```
                                               theme
                                                    |
         |--------------------------------------------------|
theme settings                                            page template settings
        |                                                                                     |
 |-------------------|-----------|                       |---------------------|------------------|
documentPage     Cover        Library          document page         cover page            library page
                                                                     template settings     template settings     template settings
```
ThemeSettings are the settings applied to new items or when the user resets its value for each template. We'll have `ThemeSettings.<template>` for each template.

But still the user can still overwrite the specific document page, cover or library page settings with template settings. 

Dis I answer your question?

packages/app-bridge/src/types/Emitter.ts
anxobotana: Because now ThemeSettings are template dependent.
We have a set of ThemeSettings per template, not a single set of theme settings.

packages/app-bridge/src/types/Emitter.ts
mike85: The requirements have changed and we need to be able to define theme settings per template.

image

henser: @mike85, with defining "theme settings per template" do that mean you can have one template defining the settings for multiple themes? or is it one template defined settings for one theme?

henser: it seems something is off to me with the diagram representation: why are `theme settings` and `page template settings` siblings? Shouldn't `theme settings` be a sibling of `template settings`? 🤔 Essentially the settings of a theme and a template should define the same properties

```
                              theme
                                |
                             settings     <- template settings plugs in here
                                |           with same underlaying child logic
                                |
                             document
                                |
           |--------------------|------------------|
        cover                  page            library 
```

Also conceptually it seems we have a domain entity naming problem: document is broader than page. A document can have multiple pages, not the other way around.

packages/app-bridge/src/types/Emitter.ts
```diff
@@ -28,7 +29,7 @@ export type EmitterEvents = {
     };
 
     'AppBridge:TemplateAssetsUpdated': {
-        template?: 'documentPage' | 'library' | 'cover';
+        template?: ThemeTemplate;
```
henser, line 32: is this really nullable? Wouldn't you need a `Template` for `**Template**AssetsUpdated` action?

packages/app-bridge/src/types/Emitter.ts
henser: "ThemeSettings are template dependent"? 🤔 Shouldn't it be the other way around? Shouldn't template depend on themes?

henser (commented):
I'm genuinely confused about the goal of template settings based on the current changes 🙈 On the owned files, code-wise I just question the nullability there. Logic-wise I'm just too far away from what the goals are there 😬

packages/app-bridge/src/types/Emitter.ts
anxobotana: @henser let's see if I can explain myself better:
Themes have `default theme settings` (for each template). Then within the guideline the editor can overwrite these `default theme settings` for each `template` those are the `Template Theme Settings`, these sets of settings are per template. These settings are apply when the editor resets or creates a new item, and they can be overwritten for that specific item.

packages/app-bridge/src/types/Emitter.ts
anxobotana: The flow of settings is as follow. We have sets of settings per template and then settings for individual items.
These settings are overwritten as the flow goes down, taking preference the latest.

```
  |                   Default Theme Settings                   <--- hardcoded in the theme
  |         (apply to all Cover, DocumentPage, Library)
 \ /                            |
  |                             |
  |                 Template Theme Settings                    <--- Editable by the Editor
  |         (apply to all Cover, DocumentPage, Library)
 \ /                            |
  |                             |
  |                     Template Settings                      <--- Editable by the Editor
  |         (individual Cover, DocumentPage or Library)
```

packages/app-bridge/src/types/Emitter.ts
anxobotana: and you are right in questioning it... fixed!

packages/app-bridge/src/types/Emitter.ts
henser: Honestly I don't see the purpose of the "Template Theme Settings", but I may be missing something. There's nothing that `Template Settings` wouldn't answer that a `Template Theme Settings` does. To me it seems that `Template Theme Settings` shouldn't be a feature itself but rather a data migration applied on `Template Settings`.

henser (approved):
I'm approving as code-wise I don't see an issue with the ecosystem team owned files but conceptually I do have concerns about the whole "Template Theme Settings". These concerns may very well stem from my ignorance on the topic 🙈 but it seems to me the whole template theme settings is a feature attempt to solve a transitional problem we need to solve internally for the Classic Theme that could be relatively easily solved with a data migration.

packages/app-bridge/src/types/Emitter.ts
anxobotana: `Template Theme Settings` makes it easier to create/reset pages/libraries with the appropriated settings with no further actions. Once these `Template Theme Settings` are set, any new page inherits them. Otherwise the editor either has to set them for every individual page/library or has to use `apply-to` every time something changes. 

`Template Settings` only applies to individual pages/libraries not to all of their kind, `Template Theme Settings` does that.

SamuelAlev (approved):

PR #678, opened 2024-01-11
feat: Add Block name and Marketplace App ID to AppBridgeBlock

utmishra:
- Adds [optional] the following optional parameters:
    - `blockName`
    - `marketplaceServiceAppId`
 - The fields will be fetched via context to be sent for event tracking payload ([Implementation PR](https://github.com/Frontify/web-app/pull/2074))


SamuelAlev:
Information shouldn't be exposed, can be extended in web-app.

PR #687, opened 2024-01-16
chore: replace deprecated functions in attachments

ragi96:
As preparation to move the `@frontify/app-bridge` to the peerDeps we need to replace all deprecated functions.


mike85 (approved):
Guidelines Navigation & Themes code looks good to me

fulopdaniel (approved):

PR #688, opened 2024-01-16
feat(Attachments): temporarily remove sorting

fulopdaniel:
Remove sorting from frontend until the endpoint for sorting assets is ready. 

https://app.clickup.com/t/8693hth4q


fulopdaniel:
Realised these are breaking changes and might go for a softer disable.

PR #689, opened 2024-01-16
feat(Attachments): temporarily disable sorting

fulopdaniel:
Temporarily disable sorting of attachments from frontend until the endpoint for sorting assets is ready.

https://app.clickup.com/t/8693hth4q


packages/guideline-blocks-settings/src/components/Attachments/types.ts
```diff
@@ -16,7 +16,7 @@ export type AttachmentsProps = {
     onDelete: (attachmentToDelete: Asset) => void;
     onUpload: (uploadedAttachments: Asset[]) => Promise<void>;
     onBrowse: (browserAttachments: Asset[]) => void;
-    onSorted: (sortedAttachments: Asset[]) => void;
+    onSorted?: (sortedAttachments: Asset[]) => void;
```
ragi96, line 19: Why this change? As it still uses the `useAttachments` Hook, its still provided always.

ragi96 (commented):
Add changeset

packages/guideline-blocks-settings/src/components/Attachments/types.ts
fulopdaniel: Initially I wanted to remove this, but realised that would be a breaking change, and since it will be reintroduced soon I did not want to go that way.
I marked is as optional because at the moment this function does not do anything.

mike85 (approved):
Guidelines Navigation & Themes code looks good to me

packages/guideline-blocks-settings/src/components/Attachments/types.ts
ragi96: Then I wouldn't mark it optionally as it would be a breaking changes if we add it later on again

ragi96 (approved):
LGTM

PR #699, opened 2024-01-19
chore: codeowners file

ragi96:
Proposal to adjust the `CODEOWNERS` to split everything to what should be owned by each team. 

Pair (programmed) with: @mike85

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,41 +1,158 @@
-* @Frontify/guidelines-navigation-themes
-**/package.json
-packages/cli/templates/**/package-lock.json
-pnpm-lock.yaml
-
-# Specific paths
-
 # App Bridge
 packages/app-bridge/src/types @Frontify/ecosystem
-packages/app-bridge/src/react/useAssetChooser.ts @Frontify/guidelines-one @Frontify/guidelines-navigation-themes @SamuelAlev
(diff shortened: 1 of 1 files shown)


julianiff (approved):

SamuelAlev (approved):

mike85 (commented):
LGTM

CODEOWNERS
```diff
@@ -1,41 +1,158 @@
...
+packages/app-bridge/src/registries/commands/TemplateChooser.ts @Frontify/guidelines-one
+
+### App Bridge registries events
+packages/app-bridge/src/registries/events/ @Frontify/guidelines-one @Frontify/guidelines-navigation-themes
+
+### App Bridge registries verbs
+packages/app-bridge/src/registries/verbs/ @Frontify/guidelines-one @Frontify/guidelines-navigation-themes @Frontify/integration
+
+## App Bridge repositories
+packages/app-bridge/src/repositories/ @Frontify/guidelines-one
```
sgruetter, line 104: @ragi96 Should all the repositories be owned just by guidelines 1? Or is this just for now and we can split it further when the time is ready?

ragi96: I think in the long run this `repositories` should be moved out of the app-bridge so we can control it within the web-app. Atm there is just one Repo there and its only used by G1, so if something gets added (i dont think so), it will be splited further

sgruetter: Ah perfect, thanks for the explanation!

PR #701, opened 2024-01-22
feat: add title to manifest.json of platform Apps

julianiff:
Adds new property to manifest.json check for Platform Apps. 
-> Now requires a title to be present.


julianiff:
clickup: https://app.clickup.com/t/8693jze29

hochreutenerl (approved):
I can't necessarily confirm that the max lengths for the title is correct (note the differences between line 10 and the others), but it should be easy to change if necessary.
I approve to unblock your work here.

PR #707, opened 2024-01-22
refactor: split testing utilities out of main bundle

SamuelAlev:
Otherwise `sinon` gets bundled into the blocks/themes/web-app


anxobotana (approved):

ragi96:
Do we need some conditional import of the testing stuff with `guideline-blocks-settings` 😅

PR #709, opened 2024-01-23
refactor: align string constants with cli

julianiff:
-> need to refactor constants to align with the manifest


imoutaharik (approved):
LGTM! 
There is just a test failing because of the Type error.

PR #719, opened 2024-01-26
fix: properly handle paths on windows

floriangaechter:
On windows, paths contain `\\` and are interpreted as [escape characters](https://github.com/mrmlnc/fast-glob#advanced-syntax) by `fastGlob`, which causes it to ignore all files. This fix replaces the `join` function with `fastGlob`s `convertPathToPattern` [method](https://github.com/mrmlnc/fast-glob#convertpathtopatternpath) where applicable.

### DIFF
--- a/.changeset/five-apples-impress.md
+++ b/.changeset/five-apples-impress.md
@@ -2,4 +2,4 @@
 "@frontify/frontify-cli": patch
 ---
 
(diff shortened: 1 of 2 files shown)


ragi96 (approved):
Just one small thing

Block deployment still works as expected

PR #720, opened 2024-01-29
fix: useAttachments use the correct functions, for adding and removing

ragi96:
https://app.clickup.com/t/8693nbz40

### DIFF
--- a/.changeset/afraid-trees-mate.md
+++ b/.changeset/afraid-trees-mate.md
@@ -1,0 +1,5 @@
+---
+"@frontify/guideline-blocks-settings": patch
+---
+
+fix(Attachments): use the according function for removing and adding asset
--- a/.changeset/afraid-trees-mate.md
+++ b/.changeset/afraid-trees-mate.md
@@ -2,4 +2,4 @@
 "@frontify/guideline-blocks-settings": patch
 ---
 
-fix(Attachments): use the according function for removing and adding asset
+fix(Attachments): use the appropriate function for removing and adding asset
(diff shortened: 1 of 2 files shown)


.changeset/afraid-trees-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/guideline-blocks-settings": patch
+---
+
+fix(Attachments): use the according function for removing and adding asset
```
SamCreasey, line 5:
```suggestion
fix(Attachments): use the appropriate function for removing and adding asset
```
Maybe?

packages/guideline-blocks-settings/src/hooks/useAttachments.tsx
```diff
@@ -6,23 +6,18 @@ import { type ReactNode, createContext, useContext } from 'react';
...
+        );
     };
 
     const onAttachmentDelete = async (assetToDelete: Asset) => {
-        const newAssetIds = attachments
-            .filter((attachment) => attachment.id !== assetToDelete.id)
-            .map((attachment) => attachment.id);
-
-        await updateAssetIdsFromKey(assetId, newAssetIds);
+        await deleteAssetIdsFromKey(assetId, [assetToDelete.id]);
```
SamCreasey, line 20: Looking at the function names should we rename the argument `assetId` in the useAttachments hook? it seems odd to call deleteAssetIdsFromKey with the first argument being `assetId`, which is actually the `key` in this case

SamCreasey (approved):
Functionally it works in the audio block, just left two minor comments that are just suggestions here

.changeset/afraid-trees-mate.md
ragi96: Seems better :)

packages/guideline-blocks-settings/src/hooks/useAttachments.tsx
ragi96: changed to `attachmentKey`

PR #722, opened 2024-01-30
feat: add label option to getBackgroundSettings

SamCreasey:
Add label and move type so that jsdoc is correctly applied for function


ragi96 (approved):
LGTM

PR #727, opened 2024-01-31
fix: cli commands brackets

ragi96:


packages/cli/src/index.ts
```diff
@@ -19,8 +19,8 @@ import pkg from '../package.json';
 
 const cli = cac(pkg.name.split('/')[1]);
 
-cli.command('login <instanceUrl>', 'log in to a Frontify instance')
-    .option('-i, --instance <instanceUrl>', '[string] url of the Frontify instance')
+cli.command('login [instanceUrl]', 'log in to a Frontify instance')
+    .option('-i, --instance [instanceUrl]', '[string] url of the Frontify instance')
```
floriangaechter, line 23: I think in this case `<instanceUrl>` is needed, because if we pass the `--instance` option, we need to pass the url as well.

packages/cli/src/index.ts
```diff
@@ -122,9 +122,9 @@ cli.command('deploy', 'deploy the app to the marketplace')
     .option('--dryRun, --dry-run', '[boolean] enable the dry run mode', { default: false })
     .option('--noVerify, --no-verify', '[boolean] disable the linting and typechecking', { default: false })
     .option('--open', '[boolean] open the marketplace app page', { default: false })
-    .option('--appType <appType>, --app-type', '[string] specify app type. Overrides manifest values')
-    .option('-i, --instance <instanceUrl>', '[string] url of the Frontify instance')
-    .option('-t, --token <accessToken>', '[string] the access token')
+    .option('--appType [appType], --app-type', '[string] specify app type. Overrides manifest values')
+    .option('-i, --instance [instanceUrl]', '[string] url of the Frontify instance')
+    .option('-t, --token [accessToken]', '[string] the access token')
```
floriangaechter, line 127: Same here. For the individual options, we need the value to be passed.

floriangaechter (approved):
LGTM 🙌

PR #731, opened 2024-02-05
feat: add theme template context to appBridgeThemes context

anxobotana:
Add theme template context to appBridgeThemes context.

### DIFF
--- a/packages/app-bridge/src/AppBridgeTheme.ts
+++ b/packages/app-bridge/src/AppBridgeTheme.ts
@@ -62,9 +62,9 @@ export type ThemeCommand = CommandNameValidator<
 >;
 
 export type TemplateContext = { templateId: string; type: ThemeTemplate } & (
-    | { type: 'documentPage'; document: Document; documentPageId: number }
+    | { type: 'documentPage'; document: Document; documentPage: DocumentPage }
     | { type: 'library'; document: Document }
-    | { type: 'cover' }
+    | { type: 'cover'; coverPage: CoverPage }
 );
 
 export type ThemeState = {
@@ -83,7 +83,7 @@ export type ThemeContext = {
     isPublicLink: boolean;
     isAuthenticated: boolean;
     languages: Language[];
-    template: TemplateContext;
+    template: TemplateContext | null;
 };
 
 export type ThemeEvent = EventNameValidator<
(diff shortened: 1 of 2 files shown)


packages/app-bridge/src/AppBridgeTheme.ts
```diff
@@ -61,6 +61,12 @@ export type ThemeCommand = CommandNameValidator<
     Pick<CommandRegistry, 'closeAssetChooser' | 'openAssetChooser' | 'openNavigationManager'>
 >;
 
+export type TemplateContext = { templateId?: string; type: ThemeTemplate } & (
```
SamuelAlev, line 64:
```suggestion
export type TemplateContext = { templateId: string; type: ThemeTemplate } & (
```
Should always be defined (right now is `default`)

packages/app-bridge/src/AppBridgeTheme.ts
```diff
@@ -61,6 +61,12 @@ export type ThemeCommand = CommandNameValidator<
...
+    | { type: 'documentPage'; document: Document; documentPageId: number }
+    | { type: 'library'; document: Document }
+    | { type: 'cover'; document: undefined }
```
SamuelAlev, line 67:
```suggestion
    | { type: 'cover'; }
```

SamuelAlev (approved):

mike85 (approved):
LGTM

packages/app-bridge/src/AppBridgeTheme.ts
```diff
@@ -77,6 +83,7 @@ export type ThemeContext = {
     isPublicLink: boolean;
     isAuthenticated: boolean;
     languages: Language[];
+    template: TemplateContext | null;
```
SamuelAlev, line 86: Why would it be `null`?

anxobotana: it can be instantiated without this data, then populate it with the template data.

SamuelAlev: Why can't you instantiate it with data?

anxobotana: Because it is `AppBridgeTheme` is instantiated in other components outside themes (`powerbar.ts` for example) or where `context.template` info is not needed. It doesn't make sense to make at least 1 api request to fulfill the template context if it's not needed.

packages/app-bridge/src/AppBridgeTheme.ts
SamuelAlev, line 1: Please add a changeset (`pnpm changeset`) with a description (with a conventional commit prefix)

anxobotana: Added

SamuelAlev (approved):

PR #732, opened 2024-02-05
feat: adjust serve and compile command

julianiff:
We would like to introduce defineApp for PlatformApps that exposes the App and a settings object. To enable that we need some changes to our deploy configuration.

### DIFF
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -6,26 +6,39 @@ import { viteExternalsPlugin } from 'vite-plugin-externals';
 import { Logger } from '../utils/logger.js';
 import { getAppBridgeVersion } from '../utils/appBridgeVersion.js';
 import pkg from '../../package.json';
+import path from 'path';
+import { fileURLToPath } from 'url';
 
 class PlatformAppDevelopmentServer {
     constructor(private readonly port: number) {}
 
     async serve(): Promise<void> {
+        Logger.info('Development');
+
+        const __dirname = path.dirname(fileURLToPath(import.meta.url));
         try {
             const viteServer = await createServer({
                 root: process.cwd(),
-                configFile: false,
-                plugins: [
-                    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-                    //@ts-ignore
-                    react(),
-                ],
+                plugins: [react()],
                 define: {
                     'process.env.NODE_ENV': JSON.stringify('development'),
                 },
-                base: `http://localhost:${this.port}/`,
+                build: {
+                    lib: {
+                        entry: path.resolve(__dirname, 'src/index.tsx'),
+                        name: 'platformApp',
+                        formats: ['es', 'umd'],
+                    },
+                    rollupOptions: {
+                        external: ['react'],
+                        output: {
+                            globals: {
+                                react: 'React',
+                            },
+                        },
+                    },
+                },
             });
-
             const server = await viteServer.listen(this.port, true);
             server.printUrls();
         } catch (error) {
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -21,7 +21,7 @@ class PlatformAppDevelopmentServer {
                 },
                 build: {
                     rollupOptions: {
-                        external: ['react', 'ReactDOM'],
+                        external: ['react', 'react-dom'],
                         output: {
                             globals: {
                                 react: 'React',
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -19,17 +19,6 @@ class PlatformAppDevelopmentServer {
                 define: {
                     'process.env.NODE_ENV': JSON.stringify('development'),
                 },
-                build: {
-                    rollupOptions: {
-                        external: ['react', 'react-dom'],
-                        output: {
-                            globals: {
-                                react: 'React',
-                                'react-dom': 'ReactDOM',
-                            },
-                        },
-                    },
-                },
             });
 
             const server = await viteServer.listen(this.port, true);
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -20,11 +20,6 @@ class PlatformAppDevelopmentServer {
                     'process.env.NODE_ENV': JSON.stringify('development'),
                 },
                 build: {
-                    lib: {
-                        entry: './src/index.tsx',
-                        name: 'platformApp',
-                        formats: ['es'],
-                    },
                     rollupOptions: {
                         external: ['react', 'ReactDOM'],
                         output: {
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -6,8 +6,6 @@ import { viteExternalsPlugin } from 'vite-plugin-externals';
 import { Logger } from '../utils/logger.js';
 import { getAppBridgeVersion } from '../utils/appBridgeVersion.js';
 import pkg from '../../package.json';
-import path from 'path';
-import { fileURLToPath } from 'url';
 
 class PlatformAppDevelopmentServer {
     constructor(private readonly port: number) {}
@@ -15,7 +13,6 @@ class PlatformAppDevelopmentServer {
     async serve(): Promise<void> {
         Logger.info('Development');
 
-        const __dirname = path.dirname(fileURLToPath(import.meta.url));
         try {
             const viteServer = await createServer({
                 root: process.cwd(),
@@ -25,20 +22,22 @@ class PlatformAppDevelopmentServer {
                 },
                 build: {
                     lib: {
-                        entry: path.resolve(__dirname, 'src/index.tsx'),
+                        entry: './src/index.tsx',
                         name: 'platformApp',
-                        formats: ['es', 'umd'],
+                        formats: ['es'],
                     },
                     rollupOptions: {
-                        external: ['react'],
+                        external: ['react', 'ReactDOM'],
                         output: {
                             globals: {
                                 react: 'React',
+                                'react-dom': 'ReactDOM',
                             },
                         },
                     },
                 },
             });
+
             const server = await viteServer.listen(this.port, true);
             server.printUrls();
         } catch (error) {
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -16,6 +16,7 @@ class PlatformAppDevelopmentServer {
         try {
             const viteServer = await createServer({
                 root: process.cwd(),
+                configFile: false,
                 plugins: [react()],
                 define: {
                     'process.env.NODE_ENV': JSON.stringify('development'),
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -11,8 +11,6 @@ class PlatformAppDevelopmentServer {
     constructor(private readonly port: number) {}
 
     async serve(): Promise<void> {
-        Logger.info('Development');
-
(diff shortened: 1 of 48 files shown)


packages/cli/src/commands/serve.ts
```diff
@@ -6,26 +6,39 @@ import { viteExternalsPlugin } from 'vite-plugin-externals';
...
 import { getAppBridgeVersion } from '../utils/appBridgeVersion.js';
 import pkg from '../../package.json';
+import path from 'path';
+import { fileURLToPath } from 'url';
 
 class PlatformAppDevelopmentServer {
     constructor(private readonly port: number) {}
 
     async serve(): Promise<void> {
+        Logger.info('Development');
```
SamuelAlev, line 16: not needed, or am I missing something?

packages/cli/src/commands/serve.ts
```diff
@@ -6,26 +6,39 @@ import { viteExternalsPlugin } from 'vite-plugin-externals';
...
-                    react(),
-                ],
+                plugins: [react()],
                 define: {
                     'process.env.NODE_ENV': JSON.stringify('development'),
                 },
-                base: `http://localhost:${this.port}/`,
+                build: {
+                    lib: {
+                        entry: path.resolve(__dirname, 'src/index.tsx'),
```
SamuelAlev, line 28:
```suggestion
                        entry: '.src/index.tsx',
```

packages/cli/src/commands/serve.ts
```diff
@@ -6,26 +6,39 @@ import { viteExternalsPlugin } from 'vite-plugin-externals';
...
+                        name: 'platformApp',
+                        formats: ['es', 'umd'],
```
SamuelAlev, line 30:
```suggestion
                        formats: ['es'],
```

packages/cli/src/commands/serve.ts
```diff
@@ -6,26 +6,39 @@ import { viteExternalsPlugin } from 'vite-plugin-externals';
...
+                    },
+                    rollupOptions: {
+                        external: ['react'],
```
SamuelAlev, line 33: Usually you should do the same for ReactDOM except if there is a reason not to

packages/cli/src/commands/serve.ts
```diff
@@ -6,26 +6,39 @@ import { viteExternalsPlugin } from 'vite-plugin-externals';
...
+                        output: {
+                            globals: {
+                                react: 'React',
+                            },
+                        },
+                    },
+                },
             });
-
             const server = await viteServer.listen(this.port, true);
```
SamuelAlev, line 42:
```suggestion

            const server = await viteServer.listen(this.port, true);
```
Let's add back a bit of air in the code

packages/cli/src/commands/serve.ts
```diff
@@ -6,26 +6,39 @@ import { viteExternalsPlugin } from 'vite-plugin-externals';
...
     constructor(private readonly port: number) {}
 
     async serve(): Promise<void> {
+        Logger.info('Development');
+
+        const __dirname = path.dirname(fileURLToPath(import.meta.url));
         try {
             const viteServer = await createServer({
                 root: process.cwd(),
-                configFile: false,
```
SamuelAlev, line 17: Removing this would allow external devs to use `vite.config.ts`, I am more inclined to have a `frontify.config.ts` which we control

packages/cli/src/utils/compiler.ts
SamuelAlev, line 1: Comments from `server.ts` applies here

packages/cli/src/commands/serve.ts
julianiff: yeah your right i dont need it anymore. When we imported the module into the index.html i didnt declare it as a module and though ill need umd. Removed it 👍

packages/cli/src/commands/serve.ts
julianiff: i got some errors before, but after retrying it all works fine

packages/cli/src/commands/serve.ts
julianiff: for context, were planning to have smth like this for our index.html

`<script type='module'>
    import platformApp from '/src/index.tsx';
    import ReactDOM from 'react-dom/client';
    import React from 'react';

    console.log('platformApp', platformApp.settings);

    const root = ReactDOM.createRoot(document.getElementById('root'));
    root.render(React.createElement(platformApp.app));
</script>`

packages/cli/src/utils/compiler.ts
```diff
@@ -80,12 +81,20 @@ export const compilePlatformApp = async ({ outputName, projectPath = '' }: Compi
     return build({
         plugins: [react(), htmlHashPlugin],
         root: projectPath,
+        define: {
+            'process.env.NODE_ENV': JSON.stringify('production'),
+        },
         build: {
             rollupOptions: {
+                external: ['react', 'ReactDOM'],
```
SamuelAlev, line 89:
```suggestion
                external: ['react', 'react-dom'],
```

packages/cli/src/commands/serve.ts
```diff
@@ -15,15 +15,21 @@ class PlatformAppDevelopmentServer {
...
-                    react(),
-                ],
+                plugins: [react()],
                 define: {
                     'process.env.NODE_ENV': JSON.stringify('development'),
                 },
-                base: `http://localhost:${this.port}/`,
+                build: {
+                    rollupOptions: {
+                        external: ['react', 'ReactDOM'],
```
SamuelAlev, line 24:
```suggestion
                        external: ['react', 'react-dom'],
```

SamuelAlev (approved):

julianiff:
@SamuelAlev Thx for your review, we realized we dont need all these changes. Ill put this PR back to draft and use it to adjust for the templates and to incorporate then the defineApp().

packages/cli/src/utils/compiler.ts
```diff
@@ -5,6 +5,7 @@ import { PluginOption, build } from 'vite';
 import { viteExternalsPlugin } from 'vite-plugin-externals';
 import { createHash } from 'crypto';
 import { getAppBridgeVersion } from './appBridgeVersion.js';
+import * as fs from 'fs';
```
SamuelAlev, line 8:
```suggestion
import { writeFileSync, readFileSync } from 'fs';
```
Use the `sync` instead of promise approach please

packages/cli/src/utils/compiler.ts
```diff
@@ -90,4 +129,13 @@ export const compilePlatformApp = async ({ outputName, projectPath = '' }: Compi
             },
         },
     });
+
+    // move and cleanup
+    try {
```
SamuelAlev, line 134: Can you explain why we have a `temp` folder?

julianiff: We are doing 2 builds. 1 that uses the index.ts as an entrypoint and creates a settings--XXXX.js file and another build just after that to create the app. Before every build the /dist folder is cleaned up, therefore we first build the settings specific build into a temp folder and then after the app is built move the settings-xxx.js to the /dist and cleanup. 

We have some javascript in our index.html, vite bundles this into the index.js file and because of that we are not able to import the {app, settings} export from the .js file directly and need to do a "lib" build to get to this file...

SamuelAlev: You can prevent that with `build.emptyOutDir` ([docs](https://vitejs.dev/config/build-options.html#build-emptyoutdir))

packages/cli/templates/platform-app-css-modules/package.json
```diff
@@ -13,8 +13,10 @@
         "@frontify/app-bridge": "^3.0.0",
         "@frontify/fondue": "12.0.0-beta.382",
         "@frontify/fondue-tokens": "^3.4.0",
+        "@frontify/platform-app": "^0.0.5",
         "react": "^18.2.0",
         "react-dom": "^18.2.0"
+
```
SamuelAlev, line 19:
```suggestion
```

packages/cli/templates/platform-app-css/package.json
```diff
@@ -13,8 +13,10 @@
         "@frontify/app-bridge": "^3.0.0",
         "@frontify/fondue": "12.0.0-beta.382",
         "@frontify/fondue-tokens": "^3.4.0",
+        "@frontify/platform-app": "^0.0.5",
         "react": "^18.2.0",
         "react-dom": "^18.2.0"
+
```
SamuelAlev, line 19:
```suggestion
```

packages/cli/templates/platform-app-tailwind/src/index.tsx
```diff
@@ -1,13 +1,19 @@
...
 
-// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-ReactDOM.createRoot(document.getElementById('root')!).render(
-    <StrictMode>
-        <App />
-    </StrictMode>,
-);
+export default defineApp({
+    app: App,
+    settings: {
```
SamuelAlev, line 9: For the templates prefer adding a `settings.ts` which `export const settings = defineSettings({ ...`

packages/cli/src/utils/compiler.ts
julianiff: oh thx, ill try that out

packages/cli/src/utils/compiler.ts
```diff
@@ -95,16 +97,26 @@ export const compilePlatformApp = async ({ outputName, entryFile, projectPath =
         build: {
             lib: {
                 entry: entryFile,
-                name: 'Settings',
-                formats: ['es'],
+                name: outputName,
+                formats: ['iife'],
```
SamuelAlev, line 99: We are looking to move to `es` when in production mode for the blocks and themes, why did you have to change for the platform apps?

packages/cli/src/utils/compiler.ts
```diff
@@ -95,16 +97,26 @@ export const compilePlatformApp = async ({ outputName, entryFile, projectPath =
...
+                fileName: () => 'index.js',
             },
             rollupOptions: {
+                external: ['app', 'react', 'react-dom'],
```
SamuelAlev, line 103: dependency called `app` doesn't exist

packages/cli/src/utils/compiler.ts
julianiff: we re-implemented how themes are calling the settings. By sticking it on the window and then directly executing allows us to access window.default.

We tried to export it as lib -> format: 'es' and using the import() to dynamically import the settings file didnt work well for us. We had cors issues and couldnt make it work..

packages/cli/src/utils/compiler.ts
```diff
@@ -77,10 +79,54 @@ export const compilePlatformApp = async ({ outputName, projectPath = '' }: Compi
         },
     };
 
-    return build({
+    const jsHashPlugin: PluginOption = {
+        name: 'js-hash',
+        enforce: 'post',
+        generateBundle(_options, bundle) {
+            bundle['index.js'].fileName = 'settings.js';
```
SamuelAlev, line 86:
```suggestion
            bundle['index.js'].fileName = 'index.js';
```
The entry is `index` not `settings` or am i wrong?

packages/cli/src/utils/compiler.ts
```diff
@@ -77,10 +79,54 @@ export const compilePlatformApp = async ({ outputName, projectPath = '' }: Compi
...
+        },
+    };
```
SamuelAlev, line 88:
```suggestion
```
as it does nothing, we can remove

SamuelAlev (approved):

mike85 (approved):
LGTM

imoutaharik (approved):

PR #734, opened 2024-02-07
feat: add new color type 

ragi96:
for supporting in guideline-blocks-settings this is needed: https://github.com/Frontify/brand-sdk/pull/744 (will also resolve failing test)

How to test:
- Use this PR web-app https://github.com/Frontify/web-app/pull/2826
- See colors in the sidebar (blocks)

Further Steps:
If this is approved merge https://github.com/Frontify/brand-sdk/pull/744 (maybe adjust it so the approved state can handle it), and sync main to dev, so ci won't fail, then merge this and release it, add new version to web-app and then its finished


packages/app-bridge/src/types/Color.ts
```diff
@@ -94,8 +94,17 @@ export type ColorApiPatch = Partial<{
 
 export type Color = {
     id: number;
-    name: Nullable<string>;
     sort: number;
+    color: ColorAsset;
```
henser, line 98: this `color` property name becomes somewhat redundant now considering that the type itself is already `Color` 🤔 wdyt?

packages/app-bridge/src/types/Color.ts
```diff
@@ -94,8 +94,17 @@ export type ColorApiPatch = Partial<{
...
+};
+
+// This type should be evolved to Asset in the long run
+export type ColorAsset = {
+    title: Nullable<string>;
+    revision: ColorRevision;
+};
```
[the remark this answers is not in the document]
henser: 🤔 I reckon not all properties of `Asset` are relevant for the different asset types (at least at the moment). Doesn't that point towards Asset being an interface rather than a type with different revision properties only? Or am I missing something? Are we moving towards having a stable Asset type where the specific type properties are then defined in these different Revision sub-types?

packages/app-bridge/src/types/Color.ts
ragi96: True I also wasn't sure, maybe something like `asset` would fit better?

packages/app-bridge/src/types/Color.ts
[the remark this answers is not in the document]
ragi96: As far I know, the asset type shouldn't have anything special which only is used for a special type of `revision`, as you only can only decide on the `payload` of the `revision` what type of an `asset` it it. Everything on the `asset` could/should be used by all assets.
Probably something we should keep in mind as it evolves in the future.

packages/app-bridge/src/types/Color.ts
henser: not sure if `asset` would be the best alternative as `Color` is already the Asset reference but maybe both `title` and `revision` could be directly referenced here ... is `ColorAsset` really necessary?

packages/app-bridge/src/types/Color.ts
```diff
@@ -134,4 +143,32 @@ export type ColorCreate = {
     alpha: number;
 };
 
-export type ColorPatch = Partial<Omit<Color, 'id' | 'hex' | 'hue' | 'saturation' | 'lightness'>>;
+export type ColorPatch = {
```
henser, line 146: Is there an advantage in defining all the color patch properties in a single object of top level properties. For example, I see how properties `c`, `m`, `y`, `k` could all be encapsulated into a `cmyk` object that would hold all cmyk related color properties (including coated, uncoated and newspaper). Same goes for pantone properties. What do you think?

henser (commented):
Let's maybe find a few minutes and sit together to discuss some of these type structures, sounds good?

packages/app-bridge/src/types/Color.ts
ragi96: ColorPatch should basically stay the same as before

packages/app-bridge/src/types/Color.ts
```diff
@@ -134,4 +127,33 @@ export type ColorCreate = {
...
+};
+
+type Cmyk = {
+    c: Nullable<number>;
+    m: Nullable<number>;
+    y: Nullable<number>;
+    k: Nullable<number>;
+    cmykCoated: Nullable<string>;
+    cmykUncoated: Nullable<string>;
+    cmykNewspaper: Nullable<string>;
```
henser, line 149: `cmyk` prefixes are no longer necessary year

packages/app-bridge/src/types/Color.ts
```diff
@@ -134,4 +127,33 @@ export type ColorCreate = {
...
+};
+
+type Pantone = {
+    pantone: Nullable<string>;
```
henser, line 153: this key becomes redundant as such: I would suggest `code` or `reference` instead

packages/app-bridge/src/types/Color.ts
```diff
@@ -134,4 +127,33 @@ export type ColorCreate = {
...
+    pantoneCoated: Nullable<string>;
+    pantoneUncoated: Nullable<string>;
+    pantoneCp: Nullable<string>;
+    pantonePlastics: Nullable<string>;
+    pantoneTextile: Nullable<string>;
```
henser, line 158: `pantone` prefixes are now redundant and should be removed  😉

packages/app-bridge/src/types/Color.ts
```diff
@@ -134,4 +127,33 @@ export type ColorCreate = {
...
+    green: Nullable<number>;
+    blue: Nullable<number>;
+    alpha: Nullable<number>;
+};
+
+type Cmyk = {
+    c: Nullable<number>;
+    m: Nullable<number>;
+    y: Nullable<number>;
+    k: Nullable<number>;
```
henser, line 146: Q: should these keys be written in extensive (cyan, magenta, yellow, black) to be consistent with Rgb?

packages/app-bridge/src/types/Color.ts
```diff
@@ -134,4 +127,33 @@ export type ColorCreate = {
     alpha: number;
 };
 
-export type ColorPatch = Partial<Omit<Color, 'id' | 'hex' | 'hue' | 'saturation' | 'lightness'>>;
+export type ColorPatch = Partial<Omit<ColorRevision, 'hex' | 'hue' | 'saturation' | 'lightness'>> & {
+    name?: Nullable<string>;
+    sort?: number;
+};
+
+type Rgba = {
```
henser, line 135: I would suffix Color to these 3 new types: `RgbaColor`, `CmykColor` and `PantoneColor`

packages/app-bridge/src/tests/ColorDummy.ts
```diff
@@ -6,108 +6,114 @@ export class ColorDummy {
...
-            cmykNewspaper: null,
-            ncs: null,
-            pantoneCp: null,
-            pantonePlastics: null,
-            pantoneTextile: null,
-            hks: null,
-            threeM: null,
-            lab: null,
+            title: 'Yellow',
+            revision: {
```
henser, line 49: should this now be adjusted to reflect the new color type encapsulation?

packages/app-bridge/src/types/Color.ts
ragi96: Yes adjusted it

packages/app-bridge/src/tests/ColorDummy.ts
ragi96: yes adjusted

packages/app-bridge/src/types/Color.ts
ragi96: removed

packages/app-bridge/src/types/Color.ts
ragi96: `code` seems good

packages/app-bridge/src/types/Color.ts
ragi96: removed 👍

packages/app-bridge/src/tests/ColorDummy.ts
```diff
@@ -6,108 +6,127 @@ export class ColorDummy {
...
+                hex: 'ff0000',
+                hue: 361,
+                saturation: 100,
+                lightness: 46,
+                rgba: { red: 255, green: 0, blue: 0, alpha: 255 },
+                cmyk: {
+                    cyan: 0,
+                    magenta: 100,
+                    yellow: 100,
+                    key: 0,
```
henser, line 22: K is for blacK, not key ... I know it's weird but I didn't come up with the idea 😅 
Please just rename key to black

packages/app-bridge/src/types/Color.ts
```diff
@@ -94,36 +96,30 @@ export type ColorApiPatch = Partial<{
...
     ncs: Nullable<string>;
-    pantoneCp: Nullable<string>;
-    pantonePlastics: Nullable<string>;
-    pantoneTextile: Nullable<string>;
     hks: Nullable<string>;
     threeM: Nullable<string>;
     lab: Nullable<string>;
+    cmyk: CmykColor;
+    rgba: RgbaColor;
+    pantone: PantoneColor;
```
henser, line 122: should these not be nullable as well?

henser (approved):
Looking much sharper now 💪 Added a couple of comments, one of them for a change, the other just a question, but approving already so I don't block you

packages/app-bridge/src/types/Color.ts
ragi96: Not nullable but its values could be null

packages/app-bridge/src/types/Color.ts
```diff
@@ -134,4 +130,33 @@ export type ColorCreate = {
     alpha: number;
 };
 
-export type ColorPatch = Partial<Omit<Color, 'id' | 'hex' | 'hue' | 'saturation' | 'lightness'>>;
+export type ColorPatch = PartialDeep<Omit<ColorRevision, 'hex' | 'hue' | 'saturation' | 'lightness'>> & {
+    name?: Nullable<string>;
```
fulopdaniel, line 134: Can this be both undefined and null? Otherwise the `?` might be redundant

fulopdaniel (approved):

PR #741, opened 2024-02-08
feat: add platformAppBridge

julianiff:
Hey,
i would like to add the platformAppBridge to use the Settings for Platform Apps.

### DIFF
--- a/packages/app-bridge/src/AppBridgePlatformApp.ts
+++ b/packages/app-bridge/src/AppBridgePlatformApp.ts
@@ -17,7 +17,7 @@ import type {
     StateReturn,
     SubscribeMap,
 } from './AppBridge';
-import { Topic } from './types';
+import { Document, DocumentPage, DocumentSection, Topic } from './types';
 import { ErrorMessageBus, IMessageBus, MessageBus } from './utilities/MessageBus';
 import { generateRandomString, notify, subscribe } from './utilities';
 import { getQueryParameters } from './utilities/queryParams';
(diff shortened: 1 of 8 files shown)


packages/app-bridge/src/AppBridgePlatformApp.ts
```diff
@@ -230,6 +230,18 @@ export class AppBridgePlatformApp implements IAppBridgePlatformApp {
...
+        return null as any;
+    }
+
+    getDocumentPagesByDocumentId(documentId: number): Promise<DocumentPage[]> {
+        return null as any;
+    }
+
+    getDocumentSectionsByDocumentPageId(documentPageId: number): Promise<DocumentSection[]> {
+        return null as any;
+    }
```
julianiff, line 244: Hey @SamuelAlev 
We have some typing issues and to fix some of them i have to add these methods. What do you think would be the best way to adapt these so we could use it for Platform Apps?

ragi96: This shouldn't be added in here, refactor the linkChooser so it displays something like `not supported for plattform apps` or just remove the internal linkSelector button if used with Plattform Apps

julianiff: What does it do?

ragi96: It's basically the linkChooser of a portal for internal links:
Screenshot 2024-02-08 at 13 30 44
And then opens this linkTree:
Screenshot 2024-02-08 at 13 31 15

SamuelAlev:
Fixed in [Frontify/web-app@`eacb0d4` (#2830)](https://github.com/Frontify/web-app/pull/2830/commits/eacb0d4ca81ca0be514a482c25b7e4c5151fac6f)

PR #756, opened 2024-02-15
feat: enable multi-type toolbar items

SamCreasey:
Remove `toolbarFlyoutItems` prop and extend the `toolbarItems` prop of the BlockItemWrapper to enable users more flexibility with toolbar items while keeping the strict typing of the previous implementation.
Users can now an array of objects, each of which should have a `type` property that is one of "dragHandle", "menu", "flyout", "button". These button object can be in any order, meaning that any item is able to open it's own flyout, be a dragHandle, open a menu or be a simple button


.changeset/olive-roses-mix.md
```diff
@@ -0,0 +1,102 @@
+---
```
ragi96, line 1: Probably we should add an example of each type

packages/guideline-blocks-settings/src/components/BlockItemWrapper/Toolbar/Toolbar.spec.tsx
```diff
@@ -33,15 +34,89 @@ describe('Toolbar', () => {
...
     });
 
     it('should throw error if toolbar does have attachments enabled without provider', () => {
-        expect(() =>
-            render(<Toolbar items={[]} flyoutMenu={{ items: [] }} attachments={{ isEnabled: true }} />),
-        ).toThrowError();
+        expect(() => render(<Toolbar items={[]} attachments={{ isEnabled: true }} />)).toThrowError();
+    });
+
+    it('should render every item type', async () => {
```
ragi96, line 44: Can we have this also tested with Keyboard? as now its only tested on click?

.changeset/olive-roses-mix.md
SamCreasey: fixed

packages/guideline-blocks-settings/src/components/BlockItemWrapper/Toolbar/Toolbar.spec.tsx
SamCreasey: fixed

ragi96: Would split the ones for keyboard to an own test

SamCreasey: fixed

ragi96 (approved):
LGTM

PR #759, opened 2024-02-16
fix: typo in fontInput.ts

anxobotana:
typo fix

CU-8693qbn4y


ragi96 (commented):
Add changeset

.changeset/long-spies-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/sidebar-settings": patch
+---
+
+Typo fix on fontInput
```
ragi96, line 5:
```suggestion
fix: typo on fontInput `aligmentChoiceValues` to `alignmentChoiceValues `
```

.changeset/long-spies-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/sidebar-settings": patch
```
ragi96, line 2: Not sure if its only a patch as its a breaking change, also not sure if we can change it that easily as if someone is using it on a Block (external devs), the existing block will break due to this

anxobotana: I had my doubts about it as well.
`@frontify/sidebar-settings` is used in web-app (to be fixed in https://github.com/Frontify/web-app/pull/2899) and in themes (to be fixed in https://github.com/Frontify/guideline-themes/pull/59)

Kenny806: Agreed. It is unfortunate, but this is a breaking change. This would best be done when we have some more breaking changes ready to be released. Until then, we will have to live with that typo

Kenny806 (approved):
Looks good with the deprecated option :+1:

ragi96 (approved):
Should work, LGTM

PR #774, opened 2024-02-21
feat: Adjust PlatformAppContext

julianiff:


packages/app-bridge/src/AppBridgePlatformApp.ts
```diff
@@ -50,24 +50,22 @@ type AppBaseProps = {
     token: string;
     marketplaceServiceAppId: string;
     connected: boolean;
+    settings: { [key: string]: string };
```
imoutaharik, line 53: Should not this be an array?

julianiff: if im not wrong the buildableSettings stores it like {"key-one": "value-one", "key-two": "value-two"} every entry in a object

imoutaharik: Yes, you're right.

imoutaharik (approved):

PR #775, opened 2024-02-22
feat: add open/close search dialog commands

mike85:
This PR adds open/close search dialog commands to the AppBridge. This will be used in themes to open/close the `web-app` search dialog.


Kenny806 (approved):
🚀

packages/app-bridge/src/types/Emitter.ts
```diff
@@ -99,6 +99,9 @@ export type EmitterEvents = {
 
     'AppBridge:OpenNavigationManager': void;
 
+    'AppBridge:OpenSearchDialog': void;
+    'AppBridge:CloseSearchDialog': void;
```
olivereisenhut, line 103: Should we follow the previous conventions and actionify the event names?
```suggestion
    'AppBridge:SearchDialogOpened': void;
    'AppBridge:SearchDialogClosed': void;
```

.changeset/giant-shrimps-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/app-bridge": minor
+---
+
+feat: add open/close search dialog commands
```
olivereisenhut, line 5: Is it worth mentioning that it's only available for the Theme Bridge?

packages/app-bridge/src/types/Emitter.ts
mike85: I would leave it in the imperative form, otherwise it would change the meaning. We don’t want to emit an event to let others know that the search has been opened/closed. We rather want to tell the `web-app` that it should open the search dialog. Same as we did with `AppBridge:OpenNavigationManager` on line 100.

olivereisenhut (approved):

PR #776, opened 2024-02-22
refactor: remove CJS and embrace ESM

SamuelAlev:
Make ESM the default way, the produced CJS are still available through a direct import `./dist/index.umd.js`


julianiff (approved):

Kenny806 (approved):
🚀

PR #780, opened 2024-02-23
feat: add open/close search dialog commands (#775)

mike85:
* feat: add open/close search dialog commands

* refactor: update emitter type

* refactor: update export order

* Create giant-shrimps-dream.md

* Update giant-shrimps-dream.md

* Update giant-shrimps-dream.md


anxobotana (approved):

packages/app-bridge/src/registries/commands/CommandRegistry.ts
```diff
@@ -20,4 +22,6 @@ export type CommandRegistry = CommandNameValidator<{
     downloadAsset: DownloadAsset;
     openNavigationManager: OpenNavigationManager;
     openNewPublication: OpenNewPublicationPayload;
+    openSearchDialog: OpenSearchDialog;
+    closeSearchDialog: CloseSearchDialog;
```
Kenny806, line 26: Wouldn't this do the same job? Having a bunch of types which are basically an alias for `void` is just adding one more step for anyone who tries to read the code.

```suggestion
    openSearchDialog: void;
    closeSearchDialog: void;
```

mike85: Thanks for your feedback. As we’re going to clean up the AppBridgeTheme very soon I followed the same pattern that we currently use for now.

Kenny806: If following a pattern introduces more work for a cleanup that is already planned, I would prioritise reducing future workload over following a pattern we're planning to scrap. If you start breaking that pattern now, it will remind us to clean up the other occurrences when we start cleaning up the AppBridgeTheme, as well as give an idea how the result should (could) look like after the cleanup to whoever works on it.

olivereisenhut (approved):
LGTM

PR #807, opened 2024-03-06
feat: add navigateToDocumentSection command to AppBridgeTheme

anxobotana:
Adds `navigateToDocumentSection` command to the AppBridgeTheme

CU-8693yma45

### DIFF
--- a/packages/app-bridge/src/registries/commands/CommandRegistry.ts
+++ b/packages/app-bridge/src/registries/commands/CommandRegistry.ts
@@ -12,6 +12,7 @@ type OpenNavigationManager = void;
 type DownloadAsset = Asset;
 type OpenSearchDialog = void;
 type CloseSearchDialog = void;
+type NavigateToDocumentSection = number | string;
 
 export type CommandRegistry = CommandNameValidator<{
     openAssetChooser?: OpenAssetChooserPayload;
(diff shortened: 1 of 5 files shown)


packages/app-bridge/src/registries/commands/CommandRegistry.ts
```diff
@@ -24,4 +25,5 @@ export type CommandRegistry = CommandNameValidator<{
     openNewPublication: OpenNewPublicationPayload;
     openSearchDialog: OpenSearchDialog;
     closeSearchDialog: CloseSearchDialog;
+    navigateToDocumentSection: NavigateToDocumentSection;
```
mike85, line 28: Wouldn’t this:

```suggestion
    navigateToDocumentPageSection: NavigateToDocumentPageSection;
```

or even this:

```suggestion
    navigateToPageSection: NavigateToPageSection;
```

be more accurate? As Documents per se don’t have sections. It’s always the pages that have sections. I would prefer the latter. Wdyt?

mike85: Ok, then it’s fine. With the domain model we’re currently working on and the AppBridgeTheme refactoring we’ll change the naming anyways. So let’s keep it for now.

PR #832, opened 2024-03-15
feat: assetInput allowDelete prop

anxobotana:
Adds prop to assetInput to allow showing the `Delete/Remove` option in the assetInputBlock


ragi96 (commented):
Looks good but changeset is missing :)

ragi96 (approved):

mike85 (approved):
LGTM

PR #834, opened 2024-03-18
feat: update useTemplateAssets to return also themeAssets

anxobotana:
`useTemplateAssets` needs to return also themeAssets so the consumer can know if the assets used are from theme or by template.

There is no way to know between these 2 cases with the current data returned, this PR helps to solve that:
- template has no asset set but there is a theme asset
- template has an asset set but there is no theme asset overriden


mike85 (approved):
LGTM

PR #841, opened 2024-03-19
feat: make useDocumentSections hook subscribe to event listeners

SamCreasey:


.changeset/shaggy-mugs-speak.md
```diff
@@ -0,0 +1,7 @@
+---
```
ragi96, line 1: Also add some words about that the `title` of the `DocumentSection` now is `nullable`

ragi96 (changes requested):
Can you also adjust the the `AppBridgeBlockStub` so it returns one with null as title and one with one set, or even more?

The guideline-blocks-settings will break as it awaits the `DocumentSections` title is not nullable

.changeset/shaggy-mugs-speak.md
anxobotana: I guess this is not a `patch`, it's a breaking change. Returning Nullable `title` most likely break themes.

ragi96: Yeah its also only added for `dev` => alpha version, but yes you are right we should mark it as `major`

(though not sure if changeset it will handle correctly and not release v5-alpha, but we will see and could adjust the changeset pr)

SamCreasey: I've added a separate changeset with the major type change to separate the updates

packages/app-bridge/src/react/useDocumentSection.ts
```diff
@@ -17,5 +62,35 @@ export const useDocumentSection = (appBridge: AppBridgeBlock | AppBridgeTheme, d
...
+            if (event.action === 'update') {
+                const { sectionId, title, slug } = event;
+                setDocumentSections((previousSections) =>
+                    updateSectionInArray(previousSections, sectionId, title, slug),
+                );
+            }
+            if (event.action === 'delete') {
+                const { sectionId } = event;
+                setDocumentSections((previousSections) => deleteSectionFromArray(previousSections, sectionId));
+            }
```
ragi96, line 85: Use a switch statement here

packages/app-bridge/src/react/useDocumentSection.spec.ts
```diff
@@ -0,0 +1,307 @@
...
+import { AppBridgeBlock, AppBridgeTheme } from '../';
+
+import { useDocumentSection } from './useDocumentSection';
+
+const DOCUMENT_PAGE_ID = 45;
+
+const documentSections = [
+    DocumentSectionDummy.withFields({ id: 464, title: null }),
+    DocumentSectionDummy.withFields({ id: 356, title: 'New Title' }),
+    DocumentSectionDummy.withFields({ id: 37675, title: ' ' }),
```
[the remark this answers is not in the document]
SamCreasey: fixed

packages/app-bridge/src/react/useDocumentSection.ts
SamCreasey: fixed

mike85 (approved):
LGTM

ragi96 (approved):

henser (approved):
Approving after discussing with @SamCreasey and assuming the naming consistency will be updated and tackled collectively in a follow-up effort👌

PR #842, opened 2024-03-20
feat: filter out null values in guideline blocks settings

SamCreasey:


packages/guideline-blocks-settings/src/components/Link/LinkSelector/LinkSelector.spec.ct.tsx
```diff
@@ -138,6 +139,38 @@ describe('Link Selector', () => {
...
+            blockId: 1,
+        });
+        (appBridge.getDocumentGroups as SinonStub) = cy.stub().returns([]);
+        (appBridge.getAllDocuments as SinonStub) = cy.stub().returns(Promise.resolve(apiDocuments));
+        (appBridge.getDocumentPagesByDocumentId as SinonStub) = cy.stub().returns(Promise.resolve(apiPages));
+        (appBridge.getDocumentSectionsByDocumentPageId as SinonStub) = cy.stub().returns(
+            Promise.resolve([
+                apiSections[0],
+                // @frontify/app-bridge v4+ supports nullable titles
+                { ...apiSections[1], title: null } as unknown as DocumentSection,
```
ragi96, line 153: We need here a new condition to check if null as title is allowed and just adding it there

ragi96: Decided to fix that in a follow up

PR #851, opened 2024-03-27
feat: restrict settings options [CU-86945y30m]

julianiff:
- Restrict Platform App Settings choices and disable access to setting choices that have links to guidelines.


imoutaharik (approved):
LGTM

PR #852, opened 2024-04-05
fix(BlockItemWrapper): hide toolbar segment if there are no items

fulopdaniel:
image

To get rid of that right border.

### DIFF
--- a/.changeset/thirty-badgers-mate.md
+++ b/.changeset/thirty-badgers-mate.md
@@ -1,0 +1,5 @@
+---
+"@frontify/guideline-blocks-settings": patch
+---
+
+feat(BlockItemWrapper): hide toolbar segment if there are no items
--- a/.changeset/thirty-badgers-mate.md
+++ b/.changeset/thirty-badgers-mate.md
@@ -2,4 +2,4 @@
 "@frontify/guideline-blocks-settings": patch
 ---
 
-feat(BlockItemWrapper): hide toolbar segment if there are no items
+fix(BlockItemWrapper): hide toolbar segment if there are no items
--- a/packages/guideline-blocks-settings/src/components/BlockItemWrapper/Toolbar/Toolbar.tsx
+++ b/packages/guideline-blocks-settings/src/components/BlockItemWrapper/Toolbar/Toolbar.tsx
@@ -18,19 +18,21 @@ export const Toolbar = ({ items, attachments }: ToolbarProps) => (
                 <AttachmentsToolbarButton />
             </ToolbarSegment>
         )}
-        <ToolbarSegment>
-            {items.map((item) => {
-                if (item.type === 'dragHandle') {
-                    return <DragHandleToolbarButton key={item.tooltip + item.type} {...item} />;
-                }
-                if (item.type === 'menu') {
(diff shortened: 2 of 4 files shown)


.changeset/thirty-badgers-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/guideline-blocks-settings": patch
+---
+
+feat(BlockItemWrapper): hide toolbar segment if there are no items
```
ragi96, line 5:
```suggestion
fix(BlockItemWrapper): hide toolbar segment if there are no items
```

Seems more like a fix to me

packages/guideline-blocks-settings/src/components/BlockItemWrapper/Toolbar/Toolbar.tsx
```diff
@@ -18,19 +18,21 @@ export const Toolbar = ({ items, attachments }: ToolbarProps) => (
...
+                    if (item.type === 'dragHandle') {
+                        return <DragHandleToolbarButton key={item.tooltip + item.type} {...item} />;
+                    }
+                    if (item.type === 'menu') {
+                        return <MenuToolbarButton key={item.tooltip + item.type} {...item} />;
+                    }
+                    if (item.type === 'flyout') {
+                        return <FlyoutToolbarButton key={item.tooltip + item.type} {...item} />;
+                    }
+                    return <ToolbarButton key={item.tooltip + item.type} {...item} />;
```
ragi96, line 33:
```suggestion
                   switch (item.type) {
                        case 'dragHandle':
                            return <DragHandleToolbarButton key={item.tooltip + item.type} {...item} />;
                        case 'menu':
                            return <MenuToolbarButton key={item.tooltip + item.type} {...item} />;
                        case 'flyout':
                            return <FlyoutToolbarButton key={item.tooltip + item.type} {...item} />;
                        default:
                            return <ToolbarButton key={item.tooltip + item.type} {...item} />;
                    }
```

Would prefer it as switch case

ragi96: Or maybe even better a toolbarButtonMap:

```typescript
const toolbarButtonMap = {
  dragHandle: DragHandleToolbarButton,
  menu: MenuToolbarButton,
  flyout: FlyoutToolbarButton,
};
``` 

and then we could do 

```typescript
{items.map((item) => {
        const ToolbarButton = toolbarButtonMap[item.type] || ToolbarButton;
        return <ToolbarButton key={`${item.tooltip}-${item.type}`} {...item} />;
})}
```

But not sure about this one

.changeset/thirty-badgers-mate.md
fulopdaniel: Adjusted

packages/guideline-blocks-settings/src/components/BlockItemWrapper/Toolbar/Toolbar.tsx
fulopdaniel: Adjusted to the switch case version. The object approach resulted in too many type errors and fixing them would cause some overhead.

ragi96 (approved):

PR #854, opened 2024-04-09
feat: add themes types

mike85:
This PR adds the types used for the navigation items for themes.


packages/app-bridge/src/types/Themes.ts
```diff
@@ -0,0 +1,67 @@
...
+export interface CoverPage {
+    id(): number;
+    title(language?: string): string;
+    isPublished(): boolean;
+}
+
+export interface DocumentGroup {
+    id(): number;
+    name(language?: string): string;
+    children(): Document[] | DocumentLibrary[] | DocumentLink[];
```
oliverschwendener, line 14:
```suggestion
    children(): (Document|DocumentLibrary|DocumentLink)[];
```

packages/app-bridge/src/types/Themes.ts
```diff
@@ -0,0 +1,67 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { LinkSettingsDisplay, LinkSettingsIconPosition } from './Document.ts';
+
+export interface CoverPage {
+    id(): number;
+    title(language?: string): string;
+    isPublished(): boolean;
```
anxobotana, line 8: there is one missing prop:
```
hiddenInNavigation(): boolean
```

the cover page, can be published but hidden in navigation.

packages/app-bridge/src/types/Themes.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export interface DocumentGroup {
+    id(): number;
+    name(language?: string): string;
```
Kenny806, line 13: Why not `title`? It would make it consistent with all the other types

packages/app-bridge/src/types/Themes.ts
```diff
@@ -0,0 +1,67 @@
...
+    slug(language?: string): string;
+    url(language?: string): string;
+    headings(): DocumentPageHeading[];
+}
+
+export interface DocumentPageLink {
+    id(): number;
+    title(language?: string): string;
+    url(): string;
+    shouldOpenInNewTab(): boolean;
```
oliverschwendener, line 60: This doesn't exist.

```suggestion
```

packages/app-bridge/src/types/Themes.ts
mike85: As agreed with @oliverschwendener we don’t need this in frontend anymore. If it is hidden, we don’t even return it from the server. This also increases the consistency across Themes.

packages/app-bridge/src/types/Themes.ts
mike85: As a DocumentGroup is something like a folder, it should be called `name`. We discussed this yesterday up and down and decided to keep it as is in the whole codebase already. Do you agree?

packages/app-bridge/src/types/Themes.ts
anxobotana: I disagree. You can have the Home page published and browsable, but you don't want it to show in the navigation bar. It can be that you want it only on the logo click or anywhere else, just don't want it on the navigation.

packages/app-bridge/src/types/Themes.ts
Kenny806: Does the user of the interface care that it is a slightly different concept? Also, this doesn't seem like a good reason for giving a different name to what is essentially the same thing - the text you display as the entity's name/title/label. I prefer when things are consistent and predictable, so that I don't have to remember there is a special case where the `title` is called `name`. Just my two cents, feel free to do with it as you please ;-)

mike85: As agreed with @oliverschwendener and @Kenny806 we’re going to change the property `name` to `title`.

packages/app-bridge/src/types/Themes.ts
SamuelAlev, line 1: it is not "theme" but rather a collection of type, could you add them to their own files and avoid any themes specific file?

mike85: Agree. However, we planned to keep them temporarily together in this "namespace" until we can get rid of the duplicate types we currently have. Can we do that once we do the AppBridge cleanup and alpha preparation? Or should we find another way to do it now?

SamuelAlev: Then `Guidelines` instead of `Themes`?

By the way, the current solution overwrite the types that are already defined (type `Document` is now present twice), so I see 2 solutions either:
- you prefix the new types 
- you go to v4 and delete the current types

mike85: Ok, then we’ll prefix the types with Guidelines for now. Thanks!

anxobotana (approved):

anxobotana (approved):

olivereisenhut (approved):

PR #860, opened 2024-04-16
feat: merge main to dev

anxobotana:
Just merging v3 changes to v4, so we can release also an updated version of v4

PR #861, opened 2024-04-16
ci: pin pnpm to 8

ragi96:
As `pnpm` 9 got released the CI breaks now as it was defined as latest, will pin it down till we update it 

To unblock: https://github.com/Frontify/brand-sdk/pull/860


anxobotana (approved):
💪

SamuelAlev (approved):

PR #862, opened 2024-04-16
feat: add secret and endpoint restrictions

julianiff:
-> adding unique constraint to keep Secret Key and Endpoint Name unique.

### DIFF
--- a/packages/cli/src/utils/verifyManifest.ts
+++ b/packages/cli/src/utils/verifyManifest.ts
@@ -54,12 +54,12 @@ export function resetEndpointNameSet() {
 }
 const endpointCallSchema = object({
     name: string().refine(
-        (id) => {
-            if (endpointNameSet.has(id)) {
+        (name) => {
+            if (endpointNameSet.has(name)) {
                 return false;
             }
 
-            endpointNameSet.add(id);
+            endpointNameSet.add(name);
             return true;
         },
         {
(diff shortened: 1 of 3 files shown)


oliverschwendener (approved):
Looks good to me, though I have no clue why `verifyManifest.ts` is owned by us 🤷

packages/cli/src/utils/verifyManifest.ts
```diff
@@ -18,11 +18,27 @@ const logoLibraryFilenameExtension = z.enum(['svg', 'jpg', 'jpeg', 'ai', 'eps',
...
+            if (secretKeySet.has(key)) {
+                return false;
+            }
+
+            secretKeySet.add(key);
+            return /^[\w-]+$/.test(key);
+        },
+        {
+            message:
+                "The key should only contain letters from a-z, A-Z, numbers from 0-9, '-' and '_' without any spaces",
```
fulopdaniel, line 39: Does it make sense to add "key must be unique" to this error message?

imoutaharik: I agree with @fulopdaniel , it would be nice to add it to the message.

julianiff: great catch updated

fulopdaniel (approved):

imoutaharik (approved):
LGTM. :)

PR #863, opened 2024-04-16
feat: add getDocumentNavigation api call

mike85:
This PR adds the getDocumentNavigation api call.

### DIFF
--- a/packages/app-bridge/src/tests/PortalNavigationTreeDummy.ts
+++ b/packages/app-bridge/src/tests/PortalNavigationTreeDummy.ts
@@ -69,7 +69,7 @@ class GuidelineDocumentGroupClass implements GuidelineDocumentGroup {
     }
 }
 
-class GuidelineDocumentClass implements GuidelineDocument {
+export class GuidelineDocumentClass implements GuidelineDocument {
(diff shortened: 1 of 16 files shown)


Kenny806 (approved):
LGTM!

packages/app-bridge/src/tests/PortalNavigationTreeDummy.ts
```diff
@@ -69,7 +69,7 @@ class GuidelineDocumentGroupClass implements GuidelineDocumentGroup {
     }
 }
 
-class GuidelineDocumentClass implements GuidelineDocument {
+export class GuidelineDocumentClass implements GuidelineDocument {
```
olivereisenhut, line 72: Interesting, should those test types be exposed? I take it on my list.

olivereisenhut (approved):

packages/app-bridge/src/tests/PortalNavigationTreeDummy.ts
mike85: In the AppBridge we export all the Dummy types for tests. Maybe we can change that in future, if not really needed.

PR #864, opened 2024-04-16
feat: merge `dev` into `main`

SamuelAlev:
Opened in draft for now to see the current state and changed files (making sure we don't ship v4 only features when not necessary)


anxobotana (approved):
LGTM, per our conversation, pending the follow up PR

anxobotana (dismissed):

PR #865, opened 2024-04-16
feat: apply changes wrongly done in v4 into v3

SamuelAlev:
v4 should only contains breaking changes of things that are already out (makes it easier to release the v4)

### DIFF
--- a/packages/app-bridge/src/types/DocumentSection.ts
+++ b/packages/app-bridge/src/types/DocumentSection.ts
@@ -7,7 +7,7 @@ export type DocumentSectionApi = {
     revision: unknown;
     slug: string;
     sort: number;
-    title: string;
+    title: Nullable<string>;
     creator: number;
     created: string;
     modifier: Nullable<number>;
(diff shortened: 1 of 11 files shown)


julianiff (approved):

packages/app-bridge/src/types/DocumentSection.ts
```diff
@@ -7,7 +7,7 @@ export type DocumentSectionApi = {
     revision: unknown;
     slug: string;
     sort: number;
-    title: string;
+    title: Nullable<string>;
```
ragi96, line 10: I would consider this as a breaking change, as with v3 you will always get `string` as title and with v4 its `string || null`

SamuelAlev: Is it used by blocks? If not, then it is an internal breaking changes as themes are not released

SamuelAlev: Link Chooser handles it already: https://github.com/Frontify/brand-sdk/blob/dev/packages/guideline-blocks-settings/src/components/Link/LinkSelector/DocumentLinks.tsx#L71
And also in the RTE then as it uses the same component

ragi96: Yes but it doesn't matter if the latest version already supports it, as also partner blocks are for sure using an older version of that. So these blocks would break. Also you never know if a partner block uses it in a different way e.g. navigation block or something like this

anxobotana (approved):
files owned look good

packages/app-bridge/src/types/DocumentSection.ts
SamuelAlev: Sounds good, that flew over my mind, thanks!

SamuelAlev: Changes reverted, so they are only present in v4

ragi96 (approved):

PR #872, opened 2024-04-17
feat: add `useDocumentNavigation` hook

oliverschwendener:
This PR adds the react hook for getting the document navigation items.

### DIFF
--- a/packages/app-bridge/src/tests/DocumentNavigationTreeDummy.ts
+++ b/packages/app-bridge/src/tests/DocumentNavigationTreeDummy.ts
@@ -91,12 +91,11 @@ class DocumentPageLinkDummy implements GuidelineDocumentPageLink {
 }
 
 class DocumentPageHeadingDummy implements GuidelineDocumentPageHeading {
-    type: 'document-page-heading';
+    readonly type = 'document-page-heading' as const;
 
     readonly #id;
 
     constructor(id: number) {
-        this.type = 'document-page-heading';
         this.#id = id;
     }
 
--- a/packages/app-bridge/src/tests/DocumentNavigationTreeDummy.ts
+++ b/packages/app-bridge/src/tests/DocumentNavigationTreeDummy.ts
@@ -9,195 +9,136 @@ import {
 } from '../types';
 
 class DocumentPageCategoryDummy implements GuidelinePageCategory {
-    static with({
-        id,
-        title,
-        slug,
-        children,
-    }: {
-        id?: number;
-        title?: string;
-        slug?: string;
-        children?: (GuidelineDocumentPage | GuidelineDocumentPageLink)[];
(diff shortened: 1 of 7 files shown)


packages/app-bridge/src/tests/DocumentNavigationTreeDummy.ts
```diff
@@ -0,0 +1,203 @@
...
+    type: 'page-category';
+
+    private constructor(
+        private readonly _id: number,
+        private readonly _title: string,
+        private readonly _slug: string,
+        private readonly _children: (GuidelineDocumentPage | GuidelineDocumentPageLink)[],
+    ) {
+        this.type = 'page-category';
+    }
```
anxobotana, line 35: Let's define these outside the constructor:

```suggestion
    readonly type = 'page-category' as const;
    readonly #id: number;
    readonly #title: string;
    readonly #slug: string;
    readonly #children: (GuidelineDocumentPage | GuidelineDocumentPageLink)[];
    
   constructor(id: number, title: string, slug: string, children: (GuidelineDocumentPage | GuidelineDocumentPageLink)[],
    ) {
        this.#id = id;
        this.#title = title;
        this.#slug = slug;
        this.#children = children;
    }
```

same thing for the others

packages/app-bridge/src/tests/DocumentNavigationTreeDummy.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    id(): number {
+        return this._id;
```
anxobotana, line 38: then you need to use:

```suggestion
        return this.#id;
```

anxobotana (changes requested):
just the changes in the dummy item

packages/app-bridge/src/tests/DocumentNavigationTreeDummy.ts
```diff
@@ -9,195 +9,136 @@ import {
...
-    }: {
-        id?: number;
-        title?: string;
-        slug?: string;
-        children?: (GuidelineDocumentPage | GuidelineDocumentPageLink)[];
-    }): GuidelinePageCategory {
-        return new DocumentPageCategoryDummy(id ?? 1234, title ?? 'Dummy Title', slug ?? 'dummy-slug', children ?? []);
-    }
-
     type: 'page-category';
```
anxobotana, line 12: if you define this here like this, you don't need to set the value in the constructor, it looks redundant:
```
readonly type = 'page-category' as const;
```

oliverschwendener: done

anxobotana (approved):
LGTM 💪

olivereisenhut (approved):

PR #875, opened 2024-04-18
feat: usePageTemplateSettings to return templateThemeSettings

anxobotana:
This is needed so we can know settings not belonging to ThemeSettings but only to pageTemplateSettings


mike85 (approved):
LGTM

PR #877, opened 2024-04-18
chore: add changeset

mike85:


.changeset/dry-jokes-bake.md
```diff
@@ -0,0 +1,5 @@
+---
```
ragi96, line 1: don't release the cli here

ragi96 (changes requested):

PR #900, opened 2024-04-29
feat: only one app bridge per iframe

julianiff:
The PlatformApp AppBridge should be only created once, but multiple times callable from child-Components. 

Added global interface on the window Object for appBridgePlatformApp.

Intended behaviour: 

```
ParentComponet: 
const appBridge = usePlatformAppBridge();

return <ChildComponent />


ChildComponent: 
// should return the same reference to have consistent state / context
const appBridge = usePlatformAppBridge();
`

### DIFF
--- a/.changeset/calm-candles-wink.md
+++ b/.changeset/calm-candles-wink.md
@@ -2,4 +2,4 @@
 "@frontify/app-bridge": patch
 ---
 
-Feat/only one app bridge per iframe
+feat(AppBridgePlatformApp): only one app bridge per iframe
--- a/.changeset/calm-candles-wink.md
+++ b/.changeset/calm-candles-wink.md
@@ -1,0 +1,5 @@
+---
+"@frontify/app-bridge": patch
+---
+
+Feat/only one app bridge per iframe
--- a/packages/app-bridge/src/AppBridgePlatformApp.ts
+++ b/packages/app-bridge/src/AppBridgePlatformApp.ts
@@ -94,6 +94,14 @@ export class AppBridgePlatformApp implements IAppBridgePlatformApp {
         'Context.connected': new Map(),
     };
 
+    constructor() {
+        if (window.appBridgePlatformApp) {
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: .changeset/calm-candles-wink.md [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/app-bridge": patch
+---
+
+Feat/only one app bridge per iframe
```
SamCreasey, line 5:
```suggestion
feat(AppBridgePlatformApp): only one app bridge per iframe
```

I would maybe add a better description to the changelog too since it is public.

packages/app-bridge/src/AppBridgePlatformApp.ts
```diff
@@ -94,6 +94,14 @@ export class AppBridgePlatformApp implements IAppBridgePlatformApp {
         'Context.connected': new Map(),
     };
 
+    constructor() {
+        if (window.appBridgePlatformApp) {
+            return window.appBridgePlatformApp;
```
SamCreasey, line 99: This is an ignorant question but I'm not knowledgable about platform apps at all. Is it possible for there to ever be two platform apps on the same page? The pr description mentions for each iframe I'm not sure what the iframe is. Also would there be any conflict with the migration to the SPA structure? For example, there may only ever be one instance of this class on the window for the whole time the user is navigating through the website, including changing pages. Is this intended?

mike85 (approved):
Codeowned file looks good.

packages/app-bridge/src/AppBridgePlatformApp.ts
julianiff: No for now we will not have more then 1 PlatformApp per iframe. That will not change anytime soon. The window will then also be the one of the iframe itself and not the global window therefore an SPA structure doesn't have changes for this one.

marguslant0 (approved):

SamCreasey (approved):

PR #909, opened 2024-05-06
feat: add new useAfterInsertion hook

SamCreasey:
Related Pr in web-app (adds properties to v3 AND v4): https://github.com/Frontify/web-app/pull/4012

### DIFF
--- a/packages/app-bridge/src/AppBridgeBlock.ts
+++ b/packages/app-bridge/src/AppBridgeBlock.ts
@@ -70,6 +70,7 @@ export type BlockContext = {
     blockId: number;
     sectionId?: number;
     isAuthenticated: boolean;
+    isNewlyInserted: boolean;
 };
 
 export type BlockEvent = EventNameValidator<
--- a/packages/app-bridge/src/AppBridgeBlock.ts
+++ b/packages/app-bridge/src/AppBridgeBlock.ts
@@ -70,7 +70,7 @@ export type BlockContext = {
     blockId: number;
     sectionId?: number;
     isAuthenticated: boolean;
-    isNewlyInserted: boolean;
+    isNew: boolean;
 };
 
 export type BlockEvent = EventNameValidator<
--- a/packages/app-bridge/src/AppBridgeBlock.ts
+++ b/packages/app-bridge/src/AppBridgeBlock.ts
@@ -70,7 +70,7 @@ export type BlockContext = {
     blockId: number;
     sectionId?: number;
     isAuthenticated: boolean;
-    isNew: boolean;
+    isNewlyInserted: boolean;
 };
 
 export type BlockEvent = EventNameValidator<
--- a/packages/app-bridge/src/react/useAfterInsertion.tsx
+++ b/packages/app-bridge/src/react/useAfterInsertion.tsx
@@ -6,11 +6,12 @@ import { type AppBridgeBlock } from '../AppBridgeBlock';
 
 export const useAfterInsertion = <T extends AppBridgeBlock>(appBridge: T, callback: () => void, enabled = true) => {
     useEffect(() => {
-        const isNew = appBridge.context('isNew').get();
-        if (enabled && isNew) {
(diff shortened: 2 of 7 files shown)


packages/app-bridge/src/react/useAfterInsertion.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import { useEffect } from 'react';
+
+import { type AppBridgeBlock } from '../AppBridgeBlock';
+
+/**
+ * Hook to execute a callback after the block has been inserted.
+ * The callback is only executed when the third argument is true (default).
+ *
+ * @example
```
ragi96, line 11: Would remove that example here and if you wanna have one add it to the changeset

packages/app-bridge/src/react/useAfterInsertion.tsx
```diff
@@ -0,0 +1,39 @@
...
+ *
+ *    useAfterInsertion(appBridge, () => buttonRef.current?.focus(), hasData);
+ *
+ *    return hasData ?
+ *      <button ref={buttonRef} onClick={() => console.log("Creating new item...")}>Create New Item</Button> :
+ *      <div>Loading...</div>;
+ * };
+ */
+
+export const useAfterInsertion = <T extends AppBridgeBlock>(appBridge: T, callback: () => void, enabled = true) => {
```
ragi96, line 30: Why is the enabled needed? Shouldn't it be fully controlled by `isNewlyInserted`? Or is there a case I don't see?

SamCreasey: As described in the comment there is a case for the 'enabled'. The primary use-case thus far for this hook is so that after inserting a block you can focus the correct first element without having to click on the block again. If the block contains any asynchronous logic this would fail. 

For example, if you have a block that will call an api and only after this request has completed will display the block content, we need to be able to wait to trigger this hook

packages/app-bridge/src/AppBridgeBlock.ts
```diff
@@ -70,6 +70,7 @@ export type BlockContext = {
     blockId: number;
     sectionId?: number;
     isAuthenticated: boolean;
+    isNewlyInserted: boolean;
```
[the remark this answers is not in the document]
SamCreasey: how about `isNew`?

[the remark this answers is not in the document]
ragi96: Well also not perfect 🤔 

As I am also not able to come up with a better name I would prefer `isNewlyInserted`

packages/app-bridge/src/react/useAfterInsertion.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+import { type AppBridgeBlock } from '../AppBridgeBlock';
+
+export const useAfterInsertion = <T extends AppBridgeBlock>(appBridge: T, callback: () => void, enabled = true) => {
+    useEffect(() => {
+        const isNew = appBridge.context('isNew').get();
+        if (enabled && isNew) {
+            callback();
+        }
+    }, [enabled, callback, appBridge]);
```
ragi96, line 13:
```suggestion
    }, [enabled, appBridge]);
```
I think we should remove the `callback` here, what do you think?

ragi96 (approved):

ragi96 (changes requested):
Sorry fat fingered the approve 😅

packages/app-bridge/src/AppBridgeBlock.ts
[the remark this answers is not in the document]
SamCreasey: fixed

packages/app-bridge/src/react/useAfterInsertion.tsx
SamCreasey: Initially wanted to keep the hook consistent with the rules of React, but since useEffectEvent is still in experimental phase let's go ahead with the removal as it shouldn't harm anything

ragi96 (approved):

PR #920, opened 2024-05-15
feat(assetViewer): add isDownloadable option to asset viewer

fulopdaniel:
This has been missing for quite a while and is a first step to solve the following tickets:
https://app.clickup.com/t/2523021/TASK-10984
https://app.clickup.com/t/8693wz874
https://app.clickup.com/t/2523021/TASK-11183

### DIFF
--- a/packages/app-bridge/src/types/Terrific.ts
+++ b/packages/app-bridge/src/types/Terrific.ts
@@ -59,3 +59,8 @@ export type AssetChooserOptions = {
           selectedValueIds?: (number | string)[];
       }
 );
+
+export type AssetViewerOptions = {
+    token: string;
+    isDownloadable?: boolean;
+};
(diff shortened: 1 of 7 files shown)


ragi96 (approved):
LGTM

packages/app-bridge/src/types/Terrific.ts
```diff
@@ -59,3 +59,8 @@ export type AssetChooserOptions = {
           selectedValueIds?: (number | string)[];
       }
 );
+
+export type AssetViewerOptions = {
+    token: string;
+    isDownloadable?: boolean;
```
olivereisenhut, line 65: Could this be merged with the recently introduced `isDownloadProtected`? Or do the two values differ in some other way?

ragi96: They differ the `isDownloadProtected` is on the asset level, here it can also be protected on the guideline level or in the block itself

olivereisenhut (approved):

PR #934, opened 2024-05-28
chore: eslintignore file

julianiff:
Add eslint ignore file

### DIFF
--- a/packages/cli/templates/platform-app-css-modules/.eslintignore
+++ b/packages/cli/templates/platform-app-css-modules/.eslintignore
@@ -1,0 +1,1 @@
+**/*.eslintrc.cjs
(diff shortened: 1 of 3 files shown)


packages/cli/templates/platform-app-css-modules/.eslintignore
SamuelAlev, line 1: What is the purpose of this?

imoutaharik: FYI @SamuelAlev That was the one and the issue wasn't happening locally, only on the PR:
https://github.com/Frontify/sdk-apps-code-examples/actions/runs/9268669009/job/25497801687

PR #935, opened 2024-05-28
fix: eslint config in platform app templates

SamuelAlev:
and update other deps


imoutaharik (commented):
Quick doubt here: I see you changed tsconfig file but tsconfig.node.json. 
Should `"moduleResolution"` not be changed to `Node`, I'm wondering.

PR #948, opened 2024-06-03
fix(RTE): on safari the first click is not recognized

fulopdaniel:
This prevents some anomalies that caused Safari to not open link toolbar on first click


ragi96 (commented):
Misses changeset

fulopdaniel:
> Misses changeset

Thanks!

ragi96 (approved):

PR #950, opened 2024-06-05
chore: remove unnecessary properties on DocumentSectionApi

hochreutenerl:
These properties are not used externally (The `DocumentSection` type should be used externally)

### DIFF
--- a/.changeset/fair-foxes-train.md
+++ b/.changeset/fair-foxes-train.md
@@ -1,0 +1,5 @@
+---
+"@frontify/app-bridge": patch
+---
+
+Remove (internally) unused properties from DocumentSectionApi
--- a/.changeset/fair-foxes-train.md
+++ b/.changeset/fair-foxes-train.md
@@ -2,4 +2,4 @@
 "@frontify/app-bridge": patch
 ---
 
-Remove (internally) unused properties from DocumentSectionApi
+fix: remove unused properties from `DocumentSectionApi`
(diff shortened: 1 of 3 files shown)


ragi96 (commented):
Misses the changeset, you can add one by `pnpm changeset`

And if we drop them here its a breaking change. Its bad that they are in here, but we can only remove them within v4.

So we can adjust the web-app that on v3 revision always is `null` and for the valid_from we can set something like timestamp 1.

And in here (v3) we only mark them deprecated

olivereisenhut (approved):

hochreutenerl:
@ragi96 I don't think they are used in v3 at all. And we don't touch the `DocumentSection` type itself

ragi96:
> And

Ah I see 🤦‍♂️ 

Yeah then please add the missing changeset, and prepare the web-app 👍

.changeset/fair-foxes-train.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/app-bridge": patch
+---
+
+Remove (internally) unused properties from DocumentSectionApi
```
ragi96, line 5: fix: remove unused properties from `DocumentSectionApi`

ragi96, line 5:
```suggestion
fix: remove unused properties from `DocumentSectionApi`
```

ragi96 (approved):

PR #952, opened 2024-06-05
fix(*): container query plugin is not a public dependency

fulopdaniel:
When using the new tailwind preset, the dependency is missing from the consumer. With this PR I aim to include `@tailwindcss/container-queries` in the bundle, so the consumer does not have to install it.

image

### DIFF
--- a/.changeset/fluffy-files-kiss.md
+++ b/.changeset/fluffy-files-kiss.md
@@ -1,0 +1,5 @@
+---
+"@frontify/guideline-blocks-settings": patch
+---
+
+fix(*): move container query plugin to a public dependency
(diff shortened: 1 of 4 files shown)


.changeset/fluffy-files-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/guideline-blocks-settings": patch
+---
+
+fix(*): move container query plugin to a public dependency
```
ragi96, line 5:
```suggestion
fix(*): move `@tailwindcss/container-queries` to a public dependency
```

ragi96 (approved):

PR #955, opened 2024-06-17
feat: add additional build step for settings

julianiff:
- Adjusts Platform App serve Command to serve settings.ts on the side

### DIFF
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -1,6 +1,8 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
 import react from '@vitejs/plugin-react';
+import * as esbuild from 'esbuild';
+import globToRegExp from 'glob-to-regexp';
 import { createServer } from 'vite';
 import { viteExternalsPlugin } from 'vite-plugin-externals';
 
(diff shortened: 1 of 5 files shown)


imoutaharik (approved):
Such a tricky one! The whole point of HMR being not to have to rebuild the whole thing but I seriously have no idea how this one could be done better. 🙉

julianiff:
> Such a tricky one! The whole point of HMR being not to have to rebuild the whole thing but I seriously have no idea how this one could be done better. 🙉

Absolutly, good point 💯 , super happy to adjust it when there are improvements.

packages/cli/src/commands/serve.ts
```diff
@@ -13,16 +15,45 @@ class PlatformAppDevelopmentServer {
...
                 define: {
                     'process.env.NODE_ENV': JSON.stringify('development'),
                 },
+                plugins: [
+                    react(),
+                    {
+                        name: 'prebuild-commands',
+                        handleHotUpdate: ({ file, server }) => {
+                            const filePattern = globToRegExp('**/settings.ts');
+                            const filePattern2 = globToRegExp('**/index.ts');
```
fulopdaniel, line 42: nit

```suggestion
                            const settingsFilePattern = globToRegExp('**/settings.ts');
                            const indexFilePattern = globToRegExp('**/index.ts');
```

fulopdaniel (approved):

packages/cli/src/commands/serve.ts
```diff
@@ -13,16 +15,45 @@ class PlatformAppDevelopmentServer {
 
     async serve(): Promise<void> {
         try {
-            const viteServer = await createServer({
+            const settingsSchema = await esbuild.context({
+                entryPoints: ['./src/index.ts'],
```
SamuelAlev, line 19: It is set on line 67
```suggestion
                entryPoints: [entryFilePath],
```

packages/cli/src/commands/serve.ts
SamuelAlev, line 42: We have the `index.ts` path set as `entryFilePath` so we can remove the regex?
For the settings, do we set it in the manifest or is it always expected to be `src/settings.ts`?
In any case, I would remove the regex if possible to have the direct path, wdyt?

julianiff: What we were struggeling with was, that we needed an output of the bundle as IIFE to then include our Settings while in dev mode (not published). 

I didnt manage to have 2 builds in vite, one that serves the App and at the same time have the index.ts built as IIFE format under a specific path. Therefore i added a 2nd build with esBuild that outputs the 2nd build. It works well. 

But, now when esbuild generates a new file, the vite server doesnt serve the output of it. It doesnt know that smth changed in the file and doesnt pick up on it. Thats why i used the hotUpdate to restart with vite and have the file available. Now of course one can put their settings wherever they want, i just help them with restarting on index.ts (if they added their settings there) or when they have them in the settings.ts file. But a dev can put these wherever they want, then the mechanism wouldnt work and they would have to restart the dev server by hand. 

Im still hoping to find a way to do it with vite, but it seems not so straight forward.

SamuelAlev: ah no no, I don't mind esbuild, just asking if we could have "direct path" rather than regex (`**/<filename>.ts`)?

```ts
handleHotUpdate: ({ file, server }) => {
    const indexFilePath = [...];
    const settinsgFilePath = [...];
    if (indexFilePath === file || settinsgFilePath === file) {
        // if the change is either in settings.ts or index.ts do a rebuild to load settings
        settingsSchema.rebuild();
        server.restart();
    }
},
```

SamuelAlev: Else we would have a conflict if any other file is called `index.ts` or `settings.ts` (in nested folders)

julianiff: ahh sure, ill change it to that, your right 👍

PR #956, opened 2024-06-17
feat: add state listener to allow push from parent

julianiff:
- Adds listener to state changes that get pushed down from the Parent. 
-> Uses token to understand message and execute on specific request. For now only 1 exists.


marguslant0 (approved):
👍

PR #967, opened 2024-06-18
fix: extend AppBridgeThemeEvent to include assetsChosen

Kenny806:


packages/app-bridge-theme/src/registries/CommandRegistry.ts
```diff
@@ -1,6 +1,6 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import { type CommandNameValidator } from 'src/types';
+import { type CommandNameValidator } from '../types/Command.ts';
```
anxobotana, line 3:
```suggestion
import { type CommandNameValidator } from '../types/Command';
```

anxobotana (approved):

PR #968, opened 2024-06-18
chore: update templates

julianiff:
update brand-sdk templates


imoutaharik (approved):
Nice! so, no more `usePlatformAppBridge` hook I see. :)
Side note: maybe we should silence the sonarcloud for code duplication on the `/templates` folder. 😅

PR #974, opened 2024-06-19
feat: templateContext and NavigationItem types

anxobotana:


packages/app-bridge-theme/src/types/Guideline.ts
```diff
@@ -11,41 +11,48 @@ export enum LinkSettingsIconPosition {
...
 
-export interface GuidelineDocumentGroup {
+interface DocumentGroupBase {
     id(): number;
     title(language?: string): string;
-    children(): (GuidelineDocument | GuidelineDocumentLibrary | GuidelineDocumentLink)[];
+    children(): (DocumentNavigationItem | DocumentLibraryNavigationItem | DocumentLinkNavigationItem)[];
     type: 'document-group';
 }
+export interface DocumentGroupNavigationItem extends DocumentGroupBase {}
```
mike85, line 31: IMO we don’t need the `DocumentGroupBase` for now. Once we need a `DocumentGroup` we can easily add it. Wdyt?

packages/app-bridge-theme/src/types/Guideline.ts
```diff
@@ -11,41 +11,48 @@ export enum LinkSettingsIconPosition {
...
 
-export interface GuidelineDocument {
+interface DocumentBase {
     id(): number;
     title(language?: string): string;
     slug(language?: string): string;
     url(language?: string): string;
     parentId(): Nullable<number>;
```
mike85, line 38: To me, it feels that `parentId()` makes only sense in the navigation tree context. However as a single `Document` I wonder whether we need it. And if we decide to keep it. Should we rename it here to `documentGroupId()`? This would help developers to better understand what this ID really refers to. Wdyt?

packages/app-bridge-theme/src/types/Guideline.ts
```diff
@@ -11,41 +11,48 @@ export enum LinkSettingsIconPosition {
...
+export interface DocumentNavigationItem extends DocumentBase {}
+export interface Document extends DocumentBase {}
 
-export interface GuidelineDocumentLibrary {
+interface DocumentLibraryBase {
     id(): number;
     title(language?: string): string;
     slug(language?: string): string;
     url(language?: string): string;
     parentId(): Nullable<number>;
```
mike85, line 49: Same question as above.

packages/app-bridge-theme/src/types/Guideline.ts
```diff
@@ -56,53 +63,59 @@ export interface GuidelineDocumentLink {
...
     type: 'document-link';
 }
+export interface DocumentLinkNavigationItem extends DocumentLinkBase {}
```
mike85, line 66: IMO we don’t need the `DocumentLinkBase` for now. Once we need a `DocumentLink` we can easily add it. Wdyt?

packages/app-bridge-theme/src/types/Guideline.ts
```diff
@@ -56,53 +63,59 @@ export interface GuidelineDocumentLink {
...
 
-export interface GuidelinePageCategory {
+interface PageCategoryBase {
     id(): number;
     title(language?: string): string;
     slug(language?: string): string;
-    children(): (GuidelineDocumentPage | GuidelineDocumentPageLink)[];
+    children(): (DocumentPageNavigationItem | DocumentPageHeadingNavigationItem)[];
```
mike85, line 72:
```suggestion
    children(): (DocumentPageNavigationItem | DocumentPageLinkNavigationItem)[];
```

packages/app-bridge-theme/src/types/Guideline.ts
```diff
@@ -56,53 +63,59 @@ export interface GuidelineDocumentLink {
...
     type: 'page-category';
 }
+export interface PageCategoryNavigationItem extends PageCategoryBase {}
```
mike85, line 75: IMO we don’t need the `PageCategoryBase` for now. Once we need a `PageCategory` we can easily add it. Wdyt?

packages/app-bridge-theme/src/types/Guideline.ts
```diff
@@ -56,53 +63,59 @@ export interface GuidelineDocumentLink {
...
+export interface DocumentPage extends DocumentPageBase {}
 
-export interface GuidelineDocumentPageLink {
+interface DocumentPageLinkBase {
     id(): number;
     title(language?: string): string;
     url(): string;
     type: 'document-page-link';
 }
+export interface DocumentPageLinkNavigationItem extends DocumentPageLinkBase {}
```
mike85, line 95: IMO we don’t need the `DocumentPageLinkBase` for now. Once we need a `DocumentPageLink` we can easily add it. Wdyt?

packages/app-bridge-theme/src/types/Guideline.ts
```diff
@@ -56,53 +63,59 @@ export interface GuidelineDocumentLink {
...
 
-export interface GuidelineDocumentPageHeading {
+interface DocumentPageHeadingBase {
     id(): number;
     title(language?: string): string;
     slug(language?: string): string;
     type: 'document-page-heading';
 }
+export interface DocumentPageHeadingNavigationItem extends DocumentPageHeadingBase {}
```
mike85, line 103: IMO we don’t need the `DocumentPageHeadingBase` for now. Once we need a `DocumentPageHeading` we can easily add it. Wdyt?

mike85 (changes requested):

packages/app-bridge-theme/src/types/Guideline.ts
anxobotana: well, we do need it. Themes already use the current `GuidelineDocumentGroup`

packages/app-bridge-theme/src/types/Guideline.ts
anxobotana: same thing that `documentGroup`

packages/app-bridge-theme/src/types/Guideline.ts
anxobotana: same thing than for `documentGroup`

packages/app-bridge-theme/src/types/Guideline.ts
anxobotana: wait, i got what you said :)

packages/app-bridge-theme/src/types/Guideline.ts
anxobotana: I think `parentId` is a good name. It gives consistency to the different types.
And in fact, i believe `DocumentPage[...]` should also have a `parentId`

packages/app-bridge-theme/src/types/Guideline.ts
mike85: @anxobotana we only use the navigation item. But my point is only that we don’t need the base interface when only one interface uses it. I would just reduce the complexity and keep it simple.

packages/app-bridge-theme/src/types/Guideline.ts
mike85: @anxobotana IMO in a tree, `parentId()` is a well established and good name and I would definitely keep it there. However, when we have a single item that is not in a tree, it would make sense that the developer that uses this interface knows which entity it refers to. I would not rename the parentId() in the NavigationItem interfaces, however I believe that it would be better to have a clear name for the normal interfaces.

anxobotana: and for the `DocumentPage[...]` (no navigation items) wouldn't it make sense also having a `documentId`?

mike85 (approved):
LGTM

PR #976, opened 2024-06-20
chore: bump fondue dependency to 12.1.8

fulopdaniel:
1. Needed to inline Fondue when testing due to some changes with the date-fns library inside fondue
2. Needed to update the plate-emoji patch as it got updated.


ragi96 (approved):

ragi96 (changes requested):
I think you also need to bump fondue within the root  `"@frontify/fondue"`

ragi96 (approved):

PR #977, opened 2024-06-20
feat(RTE): add ability to add custom class

fulopdaniel:


ragi96 (commented):
Why this change? And isn't it kinda weird if the custom class is only added in the serialized one?

fulopdaniel:
> Why this change? And isn't it kinda weird if the custom class is only added in the serialized one?

Think of it as a similar property like `column` and `gap` which is also only used in the serializer here. In the guideline-blocks repo when you initialize the RTE plugins, there you need to supply the `customClass` to the BreakAfter plugin, and Fondue will take care of the rest.

It is needed so we can use container query tailwind classnames without adding container query tailwind plugin to fondue.

PR #984, opened 2024-06-25
feat(RTE): add custom column break plugin

fulopdaniel:
Introducing container query responsivity to the rich text editor when it comes to columns.

1. Extending fondue ColumnBreak plugin with some custom logic, so guideline-blocks can stay untouched
2. Extending serialized text with the same logic

### DIFF
--- a/.changeset/spotty-papayas-admire.md
+++ b/.changeset/spotty-papayas-admire.md
@@ -3,3 +3,4 @@
 ---
 
 feat(RTE): add custom column break plugin
+Use simply as: `new BreakAfterPlugin({ columns, gap })`, and the plugin will take care of responsivity using container queries. Don't forget to also supply columns and gap to the `RichTextEditor` component.
--- a/.changeset/spotty-papayas-admire.md
+++ b/.changeset/spotty-papayas-admire.md
@@ -1,5 +1,5 @@
 ---
-"@frontify/guideline-blocks-settings": patch
+"@frontify/guideline-blocks-settings": minor
 ---
 
 feat(RTE): add custom column break plugin
--- a/.changeset/spotty-papayas-admire.md
+++ b/.changeset/spotty-papayas-admire.md
@@ -1,0 +1,5 @@
+---
+"@frontify/guideline-blocks-settings": patch
+---
+
+feat(RTE): add custom column break plugin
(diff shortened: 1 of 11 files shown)


packages/guideline-blocks-settings/src/components/Attachments/Attachments.spec.ct.tsx
```diff
@@ -103,7 +103,6 @@ describe('Attachments', () => {
         });
 
         if ((await isPre302Stub(appBridge)) && hasOpenAssetChooser(appBridge)) {
-            // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
fulopdaniel, line 106: `pnpm lint:fix` fixed this one

ragi96: Probably not, pretty sure tests will fail with AppBridge 3.0.0

fulopdaniel: yepppp...

ragi96 (commented):
Don't we need to replace the `getColumnBreakClasses` in the other Plugins?

And also add tests for the behaviour you add here, so we are sure it works as expected 💯

.changeset/spotty-papayas-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/guideline-blocks-settings": patch
+---
+
+feat(RTE): add custom column break plugin
```
ragi96, line 5: Maybe add a bit of stuff here about how the new Plugin can be used

fulopdaniel:
> Don't we need to replace the `getColumnBreakClasses` in the other Plugins?
> 
> And also add tests for the behaviour you add here, so we are sure it works as expected 💯

`getColumnBreakClasses` does not need to be replaced, it is for the dashed underline. As I am using the plugin from Fondue and just added some functionality on top, it continues to work as expected.

Added the tests.

.changeset/spotty-papayas-admire.md
```diff
@@ -0,0 +1,6 @@
+---
+"@frontify/guideline-blocks-settings": patch
```
ragi96, line 2:
```suggestion
"@frontify/guideline-blocks-settings": minor
```

ragi96 (commented):
Just one last thing

ragi96 (approved):

PR #986, opened 2024-06-26
feat: copy appBridgePlatformApp to app-bridge-app

julianiff:


imoutaharik (approved):
Looks good to me. 🎉 
I just have a quick question, I'm always confused with those 2 branches, we are merging this in main only? or dev? I see that app-bridge-theme exist only on `dev` branch. 🤔

julianiff:
> Looks good to me. 🎉 I just have a quick question, I'm always confused with those 2 branches, we are merging this in main only? or dev? I see that app-bridge-theme exist only on `dev` branch. 🤔

Dev is used to make alpha releases. But we are more interested in bringing out a 0.0.X release for the moment and then do a 0.1.0 that from then on will be supported for longer. To make alpha releases before, i dont see much value for it.

julianiff:
Like that we can then also only work on the main branch for the CLI and the bridge. That makes it also a bit simpler to handle.

imoutaharik:
> Like that we can then also only work on the main branch for the CLI and the bridge. That makes it also a bit simpler to handle.

Thanks for clarifying, makes sense to me.

PR #990, opened 2024-06-26
fix: fix linting and tests

julianiff:
-> Build was broken due to linting issues and tests


packages/app-bridge-app/src/AppBridgePlatformApp.spec.ts
```diff
@@ -88,23 +81,4 @@ describe('AppBridgePlatformApp', () => {
         });
         platformApp.dispatch(openConnection());
     });
-
-    it('should return correct object when subscribing to context', async () => {
```
imoutaharik, line 92: We are sure we don't need this anymore? :)

imoutaharik (approved):

PR #995, opened 2024-06-26
feat: add renderAppReact utility

julianiff:
Adds utility to renderApps in a react context and wait for the rendering of react until the appBridge is successfully connected. 

-> Can then be used inside of the templates to hide the implementation of the rendering.


packages/app-bridge-app/src/react/renderAppReact.spec.tsx
```diff
@@ -0,0 +1,92 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { screen, waitFor } from '@testing-library/react';
+import { describe, it, vi, beforeEach, afterEach, expect } from 'vitest';
+
+import * as AppBridge from '../AppBridgePlatformApp.ts'; // Adjust the import path
```
julianiff:
```suggestion
import * as AppBridge from '../AppBridgePlatformApp.ts';
```

packages/app-bridge-app/src/react/renderReactApp.spec.tsx
```diff
@@ -0,0 +1,92 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { screen, waitFor } from '@testing-library/react';
+import { describe, it, vi, beforeEach, afterEach, expect } from 'vitest';
+
+import * as AppBridge from '../AppBridgePlatformApp.ts'; // Adjust the import path
```
julianiff, line 6:
```suggestion
import * as AppBridge from '../AppBridgePlatformApp.ts';
```

packages/app-bridge-app/src/react/renderReactApp.ts
```diff
@@ -0,0 +1,24 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { type PlatformAppConfigExport } from '@frontify/platform-app';
+import { createElement } from 'react';
+import { createRoot } from 'react-dom/client';
+
+import { AppBridgePlatformApp } from '../AppBridgePlatformApp.ts';
+
+export const renderReactApp = (module: PlatformAppConfigExport, elementRoot = 'root') => {
```
olivereisenhut, line 9: Passing in the DOM object directly may allow for more flexible use cases in the future.

julianiff: Thats a great alternative. 
I guess open for debate what is more simple to use as the parameter is optional. Plus for the element is that react dpes it this way.

Passing the element would allow to query the dom by class/id/tag or through a query selector, one could pass the element directly and then decide himself when to connect it to the dom. Medium/Advanced use cases, maybe then they just want to build a very specific render function for themself.

It might be good in any case to rename the variable to show that it wants an elementId.

julianiff: It might also be possible to just remove the optional parameter. That would allow us to then add an optional in the future that fits the usecase best.

imoutaharik (approved):
Code  looks good to me. :)

olivereisenhut (approved):

PR #1000, opened 2024-06-27
Feat/additional prop in guideline search result dev

bojangles-m:


ragi96:
I think this is not needed anymore, if I am wrong feel free to reopen

PR #1002, opened 2024-06-28
feat: appBridge-cli include appBridgeTheme in commands

anxobotana:
CU-8694vu23v


julianiff (approved):

packages/cli/src/commands/serve.ts
```diff
@@ -66,6 +67,81 @@ class PlatformAppDevelopmentServer {
     }
 }
 
+class ThemeDevelopmentServer {
```
ragi96, line 70: Maybe we should split these `DevelopmentServer` classes out of the `serve.ts` so we can add proper Ownership on them

packages/cli/src/servers/blockDevelopmentServer.ts
ragi96, line 1: Can you add Ownership to each of these Server to the respective teams, so you can adjust them without doing too much noise

anxobotana: there you go :)

ragi96 (approved):

packages/cli/src/servers/blockDevelopmentServer.ts
ragi96: Thanks 💯

packages/cli/tests/utils/compiler.spec.ts
```diff
@@ -33,13 +33,23 @@ describe('Compiler utils', () => {
...
             await import(outputFile);
+
             expect(global.window).toHaveProperty('index');
             expect(global.window?.index.block).toBe('this is a block');
             expect(global.window?.index.settings).toMatchObject({ some: 'settings' });
             expect(global.window?.index.dependencies['@frontify/app-bridge']).toBe('^3.0.0-beta.99');
         });
     });
 
+    describe('compile Theme', () => {
```
mike85, line 44: IMO we should also test whether the dependencies and settings are attached to the window. Wdyt?

mike85 (changes requested):

packages/cli/tests/utils/compiler.spec.ts
anxobotana: done

packages/cli/tests/files/compile-theme-files/package.json
```diff
@@ -0,0 +1,9 @@
+{
+    "name": "test-theme",
+    "version": "1.0.0",
+    "main": "src/index.ts",
+    "dependencies": {
+        "@frontify/app-bridge": "^3.0.0-beta.99",
```
mike85, line 6: Do we still need this?

mike85 (approved):

packages/cli/tests/files/compile-theme-files/package.json
anxobotana: not really, deleted

PR #1013, opened 2024-07-02
refactor: remove change_ properties

hochreutenerl:
The change_ properties are being removed from the endpoints. So they can be removed from the API types too.

CU-8694y4vep


olivereisenhut (approved):
FYI: @mike85

PR #1027, opened 2024-07-10
feat: Rename the graphql .api() call to use the `execute` prefix

marguslant0:


packages/app-bridge-app/package.json
```diff
@@ -39,6 +39,9 @@
         "react": "^18",
         "react-dom": "^18"
     },
+    "dependencies": {
+        "type-fest": "^4.15.0"
```
julianiff, line 43: this is just for the types, we should be able to add it as devDependencies

packages/app-bridge-app/src/AppBridgePlatformApp.ts
```diff
@@ -16,19 +15,47 @@ import {
...
+    ? NameObject
+    : WrongNamePattern<
+          `${Exclude<Extract<keyof NameObject, string>, Extract<keyof PatternObject, string>>}`,
+          NameContext
+      >;
+
+type ApiMethodNamePattern = { [apiMethod: `${ApiVerb}${string}`]: { payload: unknown; response: unknown } };
+export type PlatformAppApiMethodNameValidator<ApiMethodNameObject> = Simplify<
+    ObjectNameValidator<ApiMethodNameObject, ApiMethodNamePattern, 'API Method'>
+>;
```
julianiff, line 48: Could you move these somewhere else? Maybe somewhere in a file with the name ApiMethodNameValidator. Might be a utilities, or types folder. 

Even tough the old appBridge had it in there, it creates a lot of noise and as it somehow all belongs together, might be easiest to put it somewhere away.

julianiff (approved):

julianiff:
🚀

PR #1051, opened 2024-07-17
feat(AppBridgeTheme): add context hooks

mike85:
https://app.clickup.com/t/869542xv6


anxobotana (approved):
LGTM
The only thing is that `useLanguage` and `useLanguages` are and will be a confusion point

mike85:
> LGTM The only thing is that `useLanguage` and `useLanguages` are and will be a confusion point

Should we split it into `useCurrentLanguage` and `useDefaultLanguage`?

anxobotana (approved):
LGTM

PR #1082, opened 2024-08-09
feat(app-bridge-app): New command `executeSecureRequest`

olivereisenhut:
This PR creates a new `executeSecureRequest` command that returns a [fetch response](https://developer.mozilla.org/en-US/docs/Web/API/Response). 
The client benefits from being able to decide what to do with the data. For example, deciding whether it's an image or not.

Since it's not possible to send complex objects with methods via `MessagePort:postMessage()`, we have to disassemble the message before sending it through the port and reassemble it on the other side.

### DIFF
--- a/packages/app-bridge-app/src/registries/api/index.ts
+++ b/packages/app-bridge-app/src/registries/api/index.ts
@@ -6,3 +6,4 @@ export * from './GetAssetResourceInformation';
 export * from './GetCurrentUser';
 export * from './GetSecureRequest';
 export * from './ExecuteGraphQl';
+export * from './ExecuteSecureRequest';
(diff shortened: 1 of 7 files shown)


packages/app-bridge-app/src/registries/api/index.ts
```diff
@@ -6,3 +6,4 @@ export * from './GetAssetResourceInformation';
 export * from './GetCurrentUser';
 export * from './GetSecureRequest';
 export * from './ExecuteGraphQl';
+export * from './ExecuteSecureRequest.ts';
```
julianiff, line 9:
```suggestion
export * from './ExecuteSecureRequest';
```

julianiff (approved):

PR #1086, opened 2024-08-28
refactor(AppBridgeTheme): use sidebar-settings

anxobotana:


packages/app-bridge-theme/package.json
```diff
@@ -40,7 +40,9 @@
     "peerDependencies": {
         "react": "^18"
     },
-    "dependencies": {},
+    "dependencies": {
```
SamuelAlev, line 43: Why is that? 🤔

SamuelAlev: Btw feel free to move that folder over here (we already have `guideline-blocks` and `platform-app` types over here)

anxobotana: that's exactly what we try to accomplish, removing`sidebar-settings` from `guideline-themes` and expose all the types from this package.

SamuelAlev: Then you need to expose the same way as in `guideline-theme/src/theme.ts` (aka set the correct generic and export)

SamuelAlev: Be aware all other app types (block and platform app) are using one app bridge **and** one package specific for the settings/helpers/utilities.

To be honest, I would move the `app-bridge-theme` to the `guideline-themes` not the other way around if I was to do it.

Cc @mike85, themes are going to far from the other app type, please align with other teams to fix it, please be careful.

mike85: Hey @SamuelAlev we decided yesterday to go the route of having only one package for the themes SDK as I aligned with @ragi96 and he also thought that at some point they’re going to refactor their AppBridgeBlock and end up having one single package as well. However, as no one knows by when they are able to do that, we just decided to stick to the current convention and create two separate packages.

The new plan would look like this:
- we create a new private repository `themes-brand-sdk` which will include `app-bridge-theme` and `guideline-themes-settings`. After the public alpha release we can then make the repo public as well.
- we remove the `guideline-themes` package from the `guideline-themes` repository and have only theme implementations in this repository. Once we handed over the Audi theme to their agency we’re going to remove it from the repository and then we can decide whether we want to open source this repository as well or keep it private.

What do you think?

SamuelAlev: > we create a new private repository themes-brand-sdk which will include app-bridge-theme and guideline-themes-settings. After the public alpha release we can then make the repo public as well.

I would just leave it here and copy/paste `/packages/guideline-themes` over here

> and have only theme implementations in this repository.
Sounds good, just like we do with [blocks](https://github.com/Frontify/guideline-blocks/) and [platform apps](https://github.com/Frontify/platform-apps/)

mike85: > > we decided yesterday to go the route of having only one package for the themes SDK as I aligned with @ragi96
> 
> Could you also ask Julian to do it then? Would be nice to have the 1 package solution for all types of apps.

Sure, will do that.

> 
> > we create a new private repository themes-brand-sdk which will include app-bridge-theme and guideline-themes-settings. After the public alpha release we can then make the repo public as well.
> 
> I would just leave it here and copy/paste `/packages/guideline-themes` over here

Ok, that’s also fine for us.

> 
> > and have only theme implementations in this repository.
> 
> Sounds good, just like we do with [blocks](https://github.com/Frontify/guideline-blocks/) and [platform apps](https://github.com/Frontify/platform-apps/)

mike85: Closing this PR now as we’ll do everything in a separate one.

PR #1108, opened 2024-09-09
feat: add scopes validator

julianiff:
Adds ZOD validator for manifest.json scopes.

### DIFF
--- a/packages/cli/src/utils/verifyManifest.ts
+++ b/packages/cli/src/utils/verifyManifest.ts
@@ -76,6 +76,28 @@ const endpointCallSchema = object({
 const hostnameRegex =
     /^(([\dA-Za-z]|[\dA-Za-z][\dA-Za-z-]*[\dA-Za-z])\.)*([\dA-Za-z]|[\dA-Za-z][\dA-Za-z-]*[\dA-Za-z])$/;
 
+const ScopeEnum = z.enum([
+    'basic:read',
+    'basic:write',
+    'account:read',
+    'webhook:read',
+    'webhook:write',
+]);
+
+const permissionsSchema = z.object({
+    permissions: z
+        .object({
+            scopes: z
+                .array(ScopeEnum)
+                .min(1, 'At least one scope is required')
+                .max(3, 'No more than 3 scopes are allowed')
+                .refine((scopes) => scopes.includes('basic:read'), {
+                    message: "'basic:read' is required in scopes",
+                }),
+        })
+        .optional(),
+});
+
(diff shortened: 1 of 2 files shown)


packages/cli/src/utils/verifyManifest.ts
```diff
@@ -76,6 +76,28 @@ const endpointCallSchema = object({
...
+const ScopeEnum = z.enum([
+    'basic:read',
+    'basic:write',
+    'account:read',
+    'webhook:read',
+    'webhook:write',
+]);
+
+const permissionsSchema = z.object({
+    permissions: z
```
imoutaharik, line 88: Interesting...why not just "scopes"? You think we'll be extending that permissions object to more than scopes?

julianiff: yeah maybe, im was not sure. Checked a bit how other frameworks do it and most have smth inbetween like permissions and there then define it further. 
Could you think of a better alternative?

packages/cli/src/utils/verifyManifest.ts
```diff
@@ -76,6 +76,28 @@ const endpointCallSchema = object({
...
+        .object({
+            scopes: z
+                .array(ScopeEnum)
+                .min(1, 'At least one scope is required')
+                .max(3, 'No more than 3 scopes are allowed')
```
imoutaharik, line 93: Only 3 are allowed? Does that includes 'basic:read'?

julianiff: ups thx, i didnt include the workflow first xD

packages/cli/src/utils/verifyManifest.ts
imoutaharik: mmmhh not sure... It's not a big deal, just seems extra. Tell me if I miss something but I don't see what we will add to that permissions object here apart from the scopes. Checking quickly what we have for example for Integrations, settings like OAuth clientId, secrets, RedirectUris etc... will be better set in Secure Requests. 🤔

julianiff: yeeh thats true, there is not so much around yet. Maybe what could be possible are some flags like, hasGuidelineAccess or in the theme of the workflows, canAccessWorkflows. In general things that we cannot easily solve through scopes but would like to have an opt-in behaviour.
Maybe more in the networking, but things like whitelisting URLs could also be interesting.

imoutaharik: Yes, You're right. There could be ways to extend this. Let's keep it as it is.

PR #1115, opened 2024-09-13
feat: deprecate dropdownBlock size prop

anxobotana:
New implementation of the dropdownBlock doesn't use `size` prop anymore

### DIFF
--- a/.changeset/itchy-months-teach.md
+++ b/.changeset/itchy-months-teach.md
@@ -2,4 +2,4 @@
 "@frontify/sidebar-settings": patch
 ---
 
-feat: deprecate dropdownBlock size prop
+refactor(dropdown): deprecate the size prop
--- a/.changeset/itchy-months-teach.md
+++ b/.changeset/itchy-months-teach.md
@@ -1,0 +1,5 @@
+---
+"@frontify/sidebar-settings": patch
+---
+
+feat: deprecate dropdownBlock size prop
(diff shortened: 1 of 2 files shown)


packages/sidebar-settings/src/blocks/dropdown.ts
```diff
@@ -22,6 +22,7 @@ export type DropdownBlock<AppBridge> = {
 
     /**
      * The size of the dropdown.
+     * @deprecated
```
ragi96, line 25:
```suggestion
     * @deprecated `size` has no effect anymore, will be removed 13.09.2025 
```

Not sure about the date convention in here, check other deprecated functions within the brand sdk

ragi96 (commented):
add changeset

packages/sidebar-settings/src/blocks/dropdown.ts
```diff
@@ -22,6 +22,7 @@ export type DropdownBlock<AppBridge> = {
 
     /**
      * The size of the dropdown.
+     * @deprecated `size` has no effect anymore, will be removed 13.09.2025
```
SamuelAlev, line 25: If it has no effect already, it can be removed sooner (like in a month)

.changeset/itchy-months-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/sidebar-settings": patch
+---
+
+feat: deprecate dropdownBlock size prop
```
ragi96, line 5:
```suggestion
refactor(dropdown): deprecate the size prop
```

packages/sidebar-settings/src/blocks/dropdown.ts
ragi96: True we can remove it here earlier, just need to make sure its still allowed within the web-app

ragi96 (approved):
🚀

PR #1122, opened 2024-09-23
feat: add guideline to verifier

julianiff:
-> To enable a surface in the library we add the guidelines as type.


imoutaharik (approved):
👍🏼

PR #1124, opened 2024-09-23
fix(*): scroll events are hijacked by dnd-kit

fulopdaniel:
Through painful trial and errors I found that the `useDndSensor` is causing the scroll to stop midway. Possibly due to the dnd-kit using some internal scroll listener, and the `useSensor` hook was recreated on every render since the second parameter was not memoised.

https://app.clickup.com/t/2523021/TASK-12175


ragi96 (approved):
LGTM

PR #1126, opened 2024-09-24
feat(*): add fondue component style css

fulopdaniel:
Otherwise it's possible that `web-app` fondue is on a newer version and does not have the appropriate CSS module styles anymore because the hash changed.

### DIFF
--- a/packages/guideline-blocks-settings/src/styles.css
+++ b/packages/guideline-blocks-settings/src/styles.css
@@ -1,3 +1,5 @@
+@import '../node_modules/@frontify/fondue/dist/packages/components/style.css';
+
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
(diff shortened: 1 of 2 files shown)


packages/guideline-blocks-settings/src/styles.css
```diff
@@ -1,3 +1,5 @@
+@import '../node_modules/@frontify/fondue/dist/packages/components/style.css';
```
ragi96, line 1: Uff can't we do that in a nicer way?

fulopdaniel: Copied from web-app 😅

I tried `@frontify/fondue-components/styles` but that did not work. Any other suggestions?

fulopdaniel: Update: `@frontify/fondue/components/styles` seems to be working. That is indeed much nicer

ragi96: Love it 👍

ragi96 (approved):

PR #1127, opened 2024-09-24
chore: Removed unused template components

findmind:
Since [removing the old template block](https://github.com/Frontify/web-app/pull/5063), these components are not needed anymore:

- `LegacyTemplateInputBlock` -> Was used in the sidebar settings of the old block, already removed from `web-app`
- ~Unused `useTemplateChooser` hook~
- ~Added deprecation notice to unused `getTemplateById` function~

To check:
- [x] Version bump / changeset needed?
- [x] Might this setting still be used by some 3rd party blocks?

https://app.clickup.com/t/8695duvh4


oliverschwendener (approved):

ragi96 (approved):

packages/app-bridge/src/react/useTemplateChooser.ts
```diff
@@ -1,27 +0,0 @@
-/* (c) Copyright Frontify Ltd., all rights reserved. */
```
ragi96, line 1: This doesn't need to be removed

ragi96 (changes requested):
And add changeset :)

ragi96 (approved):
:shipit:

PR #1141, opened 2024-10-10
fix: update to be able to use strings

julianiff:
-> Due to feedback it would allow to use secrets in the host of the resource


imoutaharik (approved):
Looks good.

PR #1144, opened 2024-10-10
chore: fix circular dep

julianiff:


olivereisenhut:
There is still a circular dependency, but now with different package versions. I'll close this PR, let's discuss further possible solutions.

PR #1149, opened 2024-10-28
fix: gitignore file on every folder level

imoutaharik:
I made the mistake to add that gitignore file on the copyFolder function which is reused.

It should be done only once in the createNewApp and at the root template directory level.

Following up and fixing bug from this PR: https://github.com/Frontify/brand-sdk/pull/1137

### DIFF
--- a/packages/cli/src/commands/create.ts
+++ b/packages/cli/src/commands/create.ts
@@ -1,11 +1,39 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import { resolve } from 'node:path';
+import path, { resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
 import pc from 'picocolors';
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/create.ts
```diff
@@ -1,11 +1,39 @@
 /* (c) Copyright Frontify Ltd., all rights reserved. */
 
-import { resolve } from 'node:path';
+import path, { resolve } from 'node:path';
```
SamuelAlev, line 3: import only `join`
```suggestion
import { join, resolve } from 'node:path';
```

imoutaharik: Fixed! thanks

marguslant0 (approved):

fulopdaniel (approved):

PR #1154, opened 2024-11-04
fix(templates): prevent delete on update without existing template ids

mikeyrayvon:
TASK-12885

This fixes an issue where a request to delete existing template ids before adding new ones returned an error if an empty array was submitted in the request parameter.


hochreutenerl (approved):
LGTM 👍

PR #1156, opened 2024-11-08
chore: change codeowners team [CU-8696jfn16]

henser:


agjacome (approved):

CODEOWNERS
```diff
@@ -1,5 +1,6 @@
 # App Bridge
-packages/app-bridge/src/types @Frontify/api-and-sdk
+
+packages/app-bridge/src/types @Frontify/frontend-platform
```
SamuelAlev, line 3: Why do we get the hot potato here? 😢

henser: 🥔 the api-and-sdk doesn't exist anymore and I reckon the FPT is the closest to the topic 😬 Would you like to change that? Who do you suggest?

sgruetter: I can provide some more context. 

While evaluating the new media delivery responsibilities and interfaces, we have concluded that the former `API and SDK`, the `File processing`, and the `Asset platform` teams become the `Media delivery` team. This is at least for the next few months, and we'll likely revisit this with progress.

With all this development, it's important to separate responsibilities, esp. things that the Media delivery team cannot do anymore. Thus, the redistribution. If the ownership (🥔) should change again, though, this can of course be arranged.

PR #1157, opened 2024-11-18
chore: gitignore improvements and fixes - [CU-8696pn7v4]

imoutaharik:
A bit of cleanup, a bit of fixes:

- Removed gitignore files from templates as it is not included anyway.
- Generates the file in the destination folder
- Separate templates by extension type.

From now on, care only to update `gitignoreTemplate.ts`.


julianiff (approved):

hochreutenerl (approved):
Looks good from the current perspective. But I am not sure if we should just remove the `.gitignore` files, but instead opt for correctly copying them when creating by a template.

imoutaharik:
> Looks good from the current perspective. But I am not sure if we should just remove the `.gitignore` files, but instead opt for correctly copying them when creating by a template.

@hochreutenerl .gitignore files are ignored by default when a package is published to npm. [See docs for reference.](https://docs.npmjs.com/cli/v9/using-npm/developers#keeping-files-out-of-your-package)

PR #1160, opened 2024-12-02
Support for automation surface and action settings

brunschgi:
The changes include:

* Manifest validation for automations (incl. Actions | Triggers with Variables)
* Support for Automation settings (by allowing to have multiple `PlatformAppSettingsStructureExport` in the settings object)

**Manifest validation**
Currently the automations surface definition has not been validated (unknowns are just bypassed). With the addition to the `verifyManifest.ts `, we make sure that the automation surface definition adheres to the schema.

**Support for Automation settings**
We iterated on the solution together with @getflourish and @olivereisenhut. The current approach is very flexible, allowing to add further settings (e.g. for project, libraries or brand-level settings) easily without the need to adjust the bundler or add additional parameters to the `defineApp`method. This makes it easier to evolve app settings in the future, which was also a wish from @SamuelAlev.

Please note: the changes to the `packages/platform-app/src/platformApps.ts` require changes in the `web-app` in order to work and provide the right default settings (e.g. in the marketplace).

### DIFF
--- a/packages/platform-app/src/platformApps.ts
+++ b/packages/platform-app/src/platformApps.ts
@@ -44,10 +44,18 @@ export type PlatformAppSimpleBlock =
 export type PlatformAppSettings = PlatformAppSimpleBlock | DynamicSettingBlock<AppBridgePlatformApp>;
 
 export type PlatformAppSettingsStructureExport = { [customSectionName: string]: PlatformAppSettings[] };
+export type PlatformAppMultiSettingsStructureExport = {
+    default: PlatformAppSettingsStructureExport;
+    automation?: {
+        actions?: {
+            [customActionId: string]: PlatformAppSettingsStructureExport;
+        };
+    };
+};
 
 export type PlatformAppConfigExport = {
     app: FC;
-    settings: PlatformAppSettingsStructureExport;
+    settings: PlatformAppMultiSettingsStructureExport | PlatformAppSettingsStructureExport;
 };
 
 /**
--- a/.changeset/nice-cups-sparkle.md
+++ b/.changeset/nice-cups-sparkle.md
@@ -1,0 +1,13 @@
+---
+"@frontify/platform-app": minor
+"@frontify/frontify-cli": patch
+---
+
+**Manifest validation**
+With the additional validation rules, we make sure that the automation surface definition adheres to the schema.
+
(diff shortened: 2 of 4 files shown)


packages/platform-app/src/platformApps.ts
```diff
@@ -44,10 +44,18 @@ export type PlatformAppSimpleBlock =
...
+        actions?: {
+            [customActionId: string]: PlatformAppSettingsStructureExport;
+        };
+    };
+};
 
 export type PlatformAppConfigExport = {
     app: FC;
-    settings: PlatformAppSettingsStructureExport;
+    settings: PlatformAppMultiSettingsStructureExport | PlatformAppSettingsStructureExport;
```
julianiff, line 58: Looks like a lot of automation specific logic, how about just doing a new Type: "automation"? was there any discussion for that? Then we wouldnt need to adjust the settings structure nor adjust the web-app implementation and could use the default settings in it. Wouldnt the only downside be that then someone would need to have 2 Apps for 2 different functionalities?

julianiff: That makes a lot of sense for sure, thanks for going into details about it. 

My follow up question is, how would you write documentation for the automations and would you make the destinctions between what an app can do and what an automation can do?

julianiff: Regarding the downsides:
1: If i understand it correctly, what is behind an Automation action is actions that correspond to the IDs in another system and not inside of an App. To enable a new action doesn't depend on the App, but on the third party system.  When we then would like to use the output of an automation in an App, we need to define somehow code that we can run in isolation somewhere. This code will not be the same as the visual Code that is used in an App. This would then create another addition to the App structure. If we would add a new type, we could use the settings for settings, and the src folder later on for executable code that we want to run somewhere. 

2 & 3 Im not sure about this, we are aiming to have the Marketplace as a lively place, to have more Apps in there sounds like the way to go.

4. Downside of one app to rule all is that for a developer its not clear what he has to do. We mix different concepts and introduce implicit rules on what goes together and what not. For example, when you build an automation you cannot use methods, or when building an App you shouldn't use the automation part in the mainfest.json because thats for the automation usecase. It makes the communication more difficult because we mix different concepts.

PR #1163, opened 2024-12-05
feat: add context instead of single contexts

julianiff:
-> uses the platformAppContext instead of the single contexts to help with typing.


imoutaharik (approved):
LGTM

PR #1165, opened 2025-01-15
fix(guidelineBlocksSettings): tailwind base

ragi96:
So the blocks don't overwrite web-app styles on that level

https://app.clickup.com/t/2523021/TASK-13317

### DIFF
--- a/.changeset/nine-bees-doubt.md
+++ b/.changeset/nine-bees-doubt.md
@@ -2,4 +2,4 @@
 "@frontify/guideline-blocks-settings": patch
 ---
 
-fix: add tailwind base for `*`, `::before` and `:affer`
+fix: add tailwind base for `*`, `::before` and `::after`
--- a/.changeset/nine-bees-doubt.md
+++ b/.changeset/nine-bees-doubt.md
@@ -1,0 +1,5 @@
+---
+"@frontify/guideline-blocks-settings": patch
+---
+
+fix: add tailwind base for `*`, `::before` and `:affer`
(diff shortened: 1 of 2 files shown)


ragi96:
> Is this essentially the same as adding
> 
> ```
> *, ::after, ::before {
>     box-sizing: content-box;
> }
> ```
> 
> in `guideline-blocks-settings/src/styles.css` ?

Basically yes.

fulopdaniel:
> > Is this essentially the same as adding
> > ```
> > *, ::after, ::before {
> >     box-sizing: content-box;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in `guideline-blocks-settings/src/styles.css` ?
> 
> Basically yes.

I would slightly prefer that as that seems a bit more maintainable to me. Also I worry a bit about possible unintended repercussions, as this will be applied to the whole application. I'd rather fix this on the issue itself. Wdyt?

ragi96:
> > > Is this essentially the same as adding
> > > ```
> > > *, ::after, ::before {
> > >     box-sizing: content-box;
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > in `guideline-blocks-settings/src/styles.css` ?
> > 
> > 
> > Basically yes.
> 
> I would slightly prefer that as that seems a bit more maintainable to me. Also I worry a bit about possible unintended repercussions, as this will be applied to the whole application. I'd rather fix this on the issue itself. Wdyt?

Wanna discuss it with you anyway

fulopdaniel (approved):
Checked together

PR #1188, opened 2025-04-10
chore: update fondue and add a new prop to RTE

peter-tudosa:


packages/guideline-blocks-settings/package.json
```diff
@@ -50,7 +50,7 @@
         "@dnd-kit/core": "^6.3.1",
         "@dnd-kit/modifiers": "^9.0.0",
         "@dnd-kit/sortable": "^10.0.0",
-        "@frontify/fondue": "^12.2.13",
+        "@frontify/fondue": "^12.9.3",
```
fulopdaniel, line 53: let's update fondue in every package, that's how we usually do it to avoid discrepancies and multiple different versions installed.

peter-tudosa: Cool! No problem!

Kenny806 (approved):

fulopdaniel (approved):

marguslant0 (approved):

PR #1205, opened 2025-04-24
feat(AppBridge): add new event for section moving

peter-tudosa:


peter-tudosa:
Frontify/web-app#8620

packages/app-bridge/src/types/DocumentSection.ts
SamuelAlev, line 1: Can you add your team as codeowner?

ragi96 (changes requested):
Do we really need to expose that event for the App Bridge Blocks? I assume we shouldn't do that.

PR #1208, opened 2025-05-09
feat: introduce state to AppBridgeTheme

Kenny806:
This will enable theme developers to store the height of any sticky elements on top of the page. The sticky elements' size is needed in web-app for properly being able to determine when a heading is in view and to be able to reliably scroll a heading to the top of the content area.

### DIFF
--- a/.changeset/strange-radios-bow.md
+++ b/.changeset/strange-radios-bow.md
@@ -1,0 +1,5 @@
+---
+"@frontify/app-bridge-theme": patch
+---
+
+Introduced a State for AppBridgeTheme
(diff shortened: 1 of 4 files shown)


.changeset/strange-radios-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/app-bridge-theme": patch
+---
+
+Introduced a State for AppBridgeTheme
```
anxobotana, line 5:
```suggestion
feat: Introduced a State for AppBridgeTheme
```

Kenny806: Thanks!

mike85 (approved):
LGTM

PR #1228, opened 2025-05-15
test: remove the flakey tests

ragi96:
remove of flakey tests, as they are failing like 80% of the time

BTW: they are already removed on the dev branch


mike85 (approved):
LGTM

PR #1274, opened 2025-08-26
feat(useAttachments): create new hook without asset call

fulopdaniel:


.changeset/five-files-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/guideline-blocks-settings": patch
+---
+
+feat(useAttachments): create new hook without asset call
```
ragi96, line 5: Can you improve the changelog here, with a small example on how to migrate? Thank you.

fulopdaniel: Sure, added

.changeset/five-files-leave.md
```diff
@@ -0,0 +1,20 @@
...
+- **Optional (advanced use)**: To avoid multiple asset calls, use the new hook directly:
+
+```ts
+import { Attachments, useAttachmentOperations } from '@frontify/guideline-blocks-settings';
+import { useBlockAssets } from '@frontify/app-bridge';
+
+const blockAssetsBundle = useBlockAssets(appBridge);
+const props = useAttachmentOperations("MY_ATTACHMENTS_KEY", blockAssetsBundle);
+
+return <Attachments {...props} />
```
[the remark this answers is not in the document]
fulopdaniel: I actually wonder if it's needed in this case. Added anyways 👍

PR #1281, opened 2025-09-02
fix: update sonarsource/sonarqube-scan-action to master

sgruetter:
This fixes issue https://github.com/SonarSource/sonarqube-scan-action/security/advisories/GHSA-f79p-9c5r-xg88

### DIFF
--- a/.github/workflows/cli-continuous-integration.yml
+++ b/.github/workflows/cli-continuous-integration.yml
@@ -46,7 +46,7 @@ jobs:
               run: pnpm --stream --filter {packages/cli} test:coverage
 
             - name: SonarCloud Scan
-              uses: sonarsource/sonarqube-scan-action@master
+              uses: sonarsource/sonarqube-scan-action@v5
               env:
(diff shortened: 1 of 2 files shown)


.github/workflows/cli-continuous-integration.yml
```diff
@@ -46,7 +46,7 @@ jobs:
               run: pnpm --stream --filter {packages/cli} test:coverage
 
             - name: SonarCloud Scan
-              uses: sonarsource/sonarqube-scan-action@v4
+              uses: sonarsource/sonarqube-scan-action@master
```
SamuelAlev, line 49: Pin to a version please, this is another attack vector to have "master"

ragi96: ah didn't know that :D

sgruetter: What's the vector @SamuelAlev? If somebody compromises their repo like with hashes for js dependencies?

PR #1294, opened 2025-09-10
feat: increase timeout of message bus

julianiff:
**What**
Increases Timeout of message bus

### DIFF
--- a/packages/app-bridge-app/src/utilities/MessageBus.ts
+++ b/packages/app-bridge-app/src/utilities/MessageBus.ts
@@ -53,7 +53,7 @@ export class MessageBus implements IMessageBus {
 
             setTimeout(() => {
                 reject(new TimeoutReachedError('operation'));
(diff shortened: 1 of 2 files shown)


packages/app-bridge-app/src/utilities/MessageBus.ts
```diff
@@ -53,7 +53,7 @@ export class MessageBus implements IMessageBus {
 
             setTimeout(() => {
                 reject(new TimeoutReachedError('operation'));
-            }, SUBSCRIBE_TIMEOUT * 10);
+            }, SUBSCRIBE_TIMEOUT * 300);
```
SamuelAlev, line 56: @julianiff @imoutaharik Let's not increase it but allow per use-cases to increase.

julianiff: Sure, lets look into it when it fits our priorities. 👍

PR #1301, opened 2025-09-18
feat: remove projectId and portalID from verifier

julianiff:


packages/cli/src/utils/verifyManifest.ts
```diff
@@ -128,8 +128,6 @@ export const platformAppManifestSchemaV1 = object({
         guideline: object({
             themeAction: object({
                 title: string().min(2).max(28),
-                projectId: string(),
-                portalId: string(),
```
imoutaharik, line 132: ah! this is no longer needed?

PR #1305, opened 2025-09-19
refactor(SidebarSettings): remove dependencies on Fondue

SamuelAlev:
Inlined the icon enum so we can remove Fondue from the package


ragi96 (approved):
NIT: 

In `choices.ts` and `inputs.ts` we are still refering to fondue for the icons, also link doesn't work at all

Screenshot 2025-09-19 at 11 14 07

PR #1319, opened 2025-10-01
feat(AppBridgeTheme): use page scroll info in context hook

anxobotana:
Creates the context with the page scroll info and hook to listen changes on it

CU-869a8ntk4

### DIFF
--- a/packages/app-bridge-theme/src/react/usePageScrollPosition.spec.ts
+++ b/packages/app-bridge-theme/src/react/usePageScrollPosition.spec.ts
@@ -1,0 +1,85 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { renderHook } from '@testing-library/react';
+import { act } from 'react';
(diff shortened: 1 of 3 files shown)


Kenny806 (commented):
If this is meant to enable reimplementing the logic of `useScrollToTopButtonVisibility` elsewhere, I wonder if the interface should be simpler. Could we simply expose:  
- `scrollTop` to enable deciding whether the go to top button should be visible
- `topOffset` to enable positioning the button
WDYT?

anxobotana:
@Kenny806 Yeah, i thought about only exposing those 2, I just left the 3 items in case someone wants to implement it in a similar way the web-app does it right now in `useScrollToTopButtonVisibility`. But I agree, most likely only those 2 should be enough. Cleaning `clientHeight`

mike85:
IMO we should only expose the bare minimum that is needed to build this feature. Wouldn’t expose more data than needed to keep the SDK surface as small as possible.

mike85 (approved):
LGTM

PR #1511, opened 2026-03-27
refactor: remove the visibility property from documents and pages

Kenny806:
Removing the `visibility` property from `Document` and `DocumentPage`. As the property is never actually returned from our APIs, it is always `undefined` and we do not have any use case for it at the moment. This also removes a lie from our type definitions :-)

### DIFF
--- a/.changeset/fifty-rice-call.md
+++ b/.changeset/fifty-rice-call.md
@@ -1,0 +1,6 @@
+---
+"@frontify/guideline-blocks-settings": patch
+"@frontify/app-bridge": patch
+---
+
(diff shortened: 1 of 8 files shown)


SamuelAlev (approved):

ragi96 (changes requested):
This is a breaking change please do it on the dev branch (we then just need to think about how we wanna map that, for v3)

ragi96:
@Kenny806 we will remove them completely in v4 after we launched the new link chooser 🙌

Kenny806:
@ragi96 Then I will just close this. Sorry about the noise

PR #1540, opened 2026-04-10
chore(cli): remove nock dependency, use vitest mocks instead

ragi96:
Replace nock HTTP interceptor with native vitest mocking:
- httpClient.spec.ts: use vi.stubGlobal('fetch', ...) to mock fetch directly
- user.spec.ts: use vi.mock to mock HttpClient module (consistent with publish.spec.ts)


mike85 (approved):
LGTM

PR #1551, opened 2026-04-16
fix(cli): Improves react and peer package version parsing.

jeremyzahner:
**Issue**

The CLI reads react, @frontify/app-bridge, and @frontify/app-bridge-theme versions from the consumer's package.json dependencies field. This breaks in pnpm workspaces that use the catalog: or workspace:* protocols, since those specifiers are passed as-is to the bundler and development servers instead of actual version numbers, making workspace-based block development unusable.

**Solution**

Version resolution now uses Node's built-in `findPackageJSON` (available since Node 22.14.0, matching the CLI's `>=22` engine requirement) to directly locate each package's `package.json` in `node_modules`. A fallback to the consumer's `package.json` is preserved for environments where `node_modules` is not yet populated, and checks `devDependencies` and `peerDependencies` in addition to `dependencies`.

**Note**

Some of the solutions here are a bit opinionated. For example wether or not devDeps and peerDeps should also be considered for the resolution logic or not would need your guys' input.

### DIFF
--- a/.changeset/calm-clocks-burn.md
+++ b/.changeset/calm-clocks-burn.md
@@ -1,0 +1,9 @@
+---
+"@frontify/frontify-cli": patch
+---
+
+fix(cli): resolve dependency versions from node_modules instead of package.json specifiers
+
+The development server and platform app compiler were reading React and App Bridge versions directly from the consumer's `package.json` `dependencies` field. This broke when using pnpm workspaces with the `catalog:` or `workspace:*` protocols, since those specifiers were passed as-is to the bundler instead of actual version numbers.
+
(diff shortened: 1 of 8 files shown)


jeremyzahner:
Reviewed by @lzurbriggen internally.

SamuelAlev (approved):
Pretty cool, thank you for the fix!
Also works with `link:<path>`

jeremyzahner:
> Pretty cool, thank you for the fix! Also works with `link:<path>`

That was a nice side effect that helped in testing as well. =)

jeremyzahner:
Checking the test failure.

jeremyzahner:
@SamuelAlev Pretty sure the CI failure is unrelated to my change.

ragi96 (approved):
Thank you for your contribution 🚀

ragi96:
> @SamuelAlev Pretty sure the CI failure is unrelated to my change.

Yep, unrelated. Will merge :)

PR #1559, opened 2026-04-22
Improves deployment logic for handling alternate dependency protocols.

jeremyzahner:
Hey guys!

Here my follow up from the last PR. When starting to use the new SDK revision, we noticed we we're still having runtime issues with the blocks after deployment (using the `catalog` protocol in our package.json). My last fix solved the issue when using the dev server, but not for deployments. This should solve that aspect as well. ~~For now it's a draft since i wanted to speak to @floriangaechter before marking it ready.~~

Edit 1 - ~~One question: For better observability I've added logging for resolved deps, source files and dist files when deploying. I think its generally useful for users deploying to have more transparency. However i see that this is somewhat opinionated, so let me know if I should remove it.~~ -> Removed the verbose logs.

Edit 2 - Also as a remark: As per other setups even today, this does not fix the issue of not all source files (ie. shared monorepo libs) being deployed to enable easier reviews. This is simply out of scope for this MR. I understand that this is an ongoing discussion. These changes however enable working deployments out of workspace/catalog monorepos without user needing to implement workarounds for using default pnpm features.

---
"@frontify/frontify-cli": patch
---

fix(cli): resolve workspace protocol specifiers during deployment

The deploy command now resolves `catalog:`, `workspace:`, `link:`, `file:`, and other protocol specifiers to their actual installed versions before uploading to the Frontify Marketplace API. Previously, these raw specifiers were sent as-is, causing deployment failures in pnpm workspace setups.

Key changes:

- Dependencies are resolved via `node_modules` lookup during `collectFiles`. Unresolvable protocol specifiers are omitted from the payload with a warning.
- The `package.json` included in `source_files` is sanitized with resolved versions before upload.
- The platform app compiler now guards against protocol specifiers being injected into compiled JavaScript output.
- A warning is emitted when potentially sensitive files (`.env`, `.npmrc`, `.netrc`) are detected in the source upload.

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -19,6 +19,7 @@ import {
     readFileAsBase64,
     readFileLinesAsArray,
 } from '../utils/index';
+import { isPackageProtocolSpecifier } from '../utils/packageProtocols';
 import { platformAppManifestSchemaV1, verifyManifest } from '../utils/verifyManifest';
 
 type Options = {
@@ -64,12 +65,6 @@ const SOURCE_FILE_BLOCK_LIST = [
 
 const SENSITIVE_FILE_PATTERNS = ['.env', '.npmrc', '.netrc'];
 
-const PROTOCOL_PREFIXES = ['catalog:', 'workspace:', 'link:', 'file:', 'portal:'];
-
-const isProtocolSpecifier = (specifier: string): boolean => {
-    return PROTOCOL_PREFIXES.some((prefix) => specifier.startsWith(prefix));
-};
-
 export const resolveDependencyVersions = (
     dependencies: Record<string, string>,
     projectPath: string,
@@ -79,12 +74,12 @@ export const resolveDependencyVersions = (
     for (const [name, specifier] of Object.entries(dependencies)) {
         const installedVersion = getInstalledPackageVersion(projectPath, name);
 
-        if (installedVersion && !isProtocolSpecifier(installedVersion)) {
+        if (installedVersion && !isPackageProtocolSpecifier(installedVersion)) {
             resolved[name] = installedVersion;
             continue;
         }
 
-        if (isProtocolSpecifier(specifier)) {
+        if (isPackageProtocolSpecifier(specifier)) {
             Logger.warn(
                 `Could not resolve version for "${name}" (specifier: "${specifier}"). ` +
                     'The package may not be installed. Omitting from deployment dependencies.',
@@ -173,14 +168,22 @@ export const collectFiles = async (projectPath: string, distPath: string) => {
     // Note: packageJsonContent is a reactiveJson Proxy. We only READ from it here
     // via spread and property access. Do not mutate it directly — the Proxy's set
     // trap would write changes back to disk.
-    if (sourceFiles['/package.json'] && packageJsonContent) {
+    const pkgJsonKey = '/package.json';
+    if (packageJsonContent) {
+        if (!(pkgJsonKey in sourceFiles)) {
+            Logger.error(
+                `Expected "${pkgJsonKey}" in sourceFiles but key was not found. ` +
+                    'This likely indicates a change in makeFilesDict key format.',
+            );
+            process.exit(-1);
+        }
         const sanitized = {
             ...packageJsonContent,
             dependencies: resolvedDependencies,
             devDependencies: resolveDependencyVersions(packageJsonContent.devDependencies || {}, projectPath),
             peerDependencies: resolveDependencyVersions(packageJsonContent.peerDependencies || {}, projectPath),
         };
-        sourceFiles['/package.json'] = Buffer.from(JSON.stringify(sanitized, null, '\t')).toString('base64');
+        sourceFiles[pkgJsonKey] = Buffer.from(JSON.stringify(sanitized, null, '\t')).toString('base64');
     }
 
     warnAboutSensitiveFiles(sourceFiles);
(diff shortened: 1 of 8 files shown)


ragi96:
@jeremyzahner please sign your commits :)

jeremyzahner:
> @jeremyzahner please sign your commits :) 

@ragi96  Sorry, did press the Rebase Button on GH. Will do asap.

packages/cli/src/commands/deploy.ts
```diff
@@ -108,17 +158,68 @@ export const collectFiles = async (projectPath: string, distPath: string) => {
...
+    }
+
+    warnAboutSensitiveFiles(sourceFiles);
+
+    const buildFiles = await makeFilesDict(
+        fastGlob.convertPathToPattern(`${projectPath}/${distPath}`),
+        buildFilesToIgnore,
     );
 
+    Logger.info('Resolved dependencies:');
```
mike85: @jeremyzahner I’d lean towards removing these logs. They add quite a bit of noise to the console without bringing much value for most developers, since they don’t really need insight into these internals.
If there’s still a need for this kind of visibility, we could always introduce a `--verbose` (or debug) mode later where such logs are more intentional and helpful.
Would that approach work for you?

packages/cli/src/utils/compiler/compilePlatformApp.ts
```diff
@@ -8,8 +8,15 @@ import { getAppBridgeVersion } from '../getPackageVersion';
 
 import { type CompilerOptions } from './compilerOptions';
 
+const PROTOCOL_PREFIXES = ['catalog:', 'workspace:', 'link:', 'file:', 'portal:'];
```
ragi96: This one is doubled as it as well exists here: `deploy.ts:65
`
We should move it outside so we only have to maintain one `PROTOCOL_PREFIXES`

packages/cli/src/utils/logger.ts
```diff
@@ -21,6 +21,10 @@ export class Logger {
         console.error(pc.red(`[${getCurrentTime()}] ${messages.join(' ')}`));
     }
 
+    static warn(...messages: string[]): void {
```
ragi96, line 24: This one should only be logged behind a `--verbose` flag

Feel free to create a new PR to add this mode 👍

packages/cli/src/commands/deploy.ts
```diff
@@ -108,17 +158,68 @@ export const collectFiles = async (projectPath: string, distPath: string) => {
...
+    const resolvedDependencies = resolveDependencyVersions(packageJsonContent?.dependencies || {}, projectPath);
+
+    const sourceFiles = await makeFilesDict(fastGlob.convertPathToPattern(projectPath), sourceFilesToIgnore);
+
+    // Sanitize the package.json in source_files to contain resolved versions
+    // instead of protocol specifiers that are invalid outside the workspace.
+    // Note: packageJsonContent is a reactiveJson Proxy. We only READ from it here
+    // via spread and property access. Do not mutate it directly — the Proxy's set
+    // trap would write changes back to disk.
+    if (sourceFiles['/package.json'] && packageJsonContent) {
```
ragi96: Fragile key coupling: checks `sourceFiles['/package.json']`. The leading `/` comes from `makeFilesDict`'s internal convention. If that ever changes, the sanitization silently no-ops and raw `catalog:` specifiers leak back into the uploaded `source_files/package.json` — defeating half the fix. Either extract the prefix into a shared constant alongside `makeFilesDict`, or assert/throw if the key is missing when `packageJsonContent` is non-null.

mike85 (changes requested):

packages/cli/src/commands/deploy.ts
jeremyzahner: Will remove them for now. Maybe set up a separate PR for this later on.

packages/cli/src/utils/logger.ts
jeremyzahner: @ragi96 Looking at how i used it right now (for example warning about potentially sensitive files being leaked) I'm not sure I would see it as being a "verbose" case. I would treat this as a "warning" much rather than as a "debug" message of sorts which should absolutely be placed behind a flag like this (like the verbose logs i removed about the source file list ie.). Do you want me to remove this entirely for now?

packages/cli/src/utils/compiler/compilePlatformApp.ts
jeremyzahner: @ragi96 Good catch, thanks. Obviously not maintainable like this. Will fix and move it into utils as packageProtocols.ts for example.

packages/cli/src/commands/deploy.ts
jeremyzahner: @ragi96 Good points. I was thinking this should not be a major concern since the function (makeFilesDict) is private to this module and the coupling is only internal. Also, breaking with the leading slash convention would also break so many other things (since the API needs the keys in this format). Also, tried to cover it in the tests so they would also break in case of potentially problematic changess.

Anyhow, to address this without adding an unreasonable amount of complexity i would opt for the defensive throw as suggested by you.

jeremyzahner: @ragi96 Fixed in 774f8da96f713b6aff8501d94c53eba2e9f98b42

packages/cli/src/utils/compiler/compilePlatformApp.ts
jeremyzahner: @ragi96 Fixed in 774f8da96f713b6aff8501d94c53eba2e9f98b42

packages/cli/src/commands/deploy.ts
jeremyzahner: @mike85 Removed in ba8fefa882db1fe01534406bbfb991911e2d9e73.

jeremyzahner:
@mike85  @ragi96 Not sure how you usually proceed with the conversations (if I should resolve them or you will) so I let them open for your reference.

PR #1564, opened 2026-04-24
feat(CLI): validate manifest command

ragi96:
Needs: https://github.com/Frontify/app-server/pull/30928

Adds a command to verify a manifest against the server @Frontify/developer-experience feel free to move your FE check also to the BE ;) 

Also verifys the manifest before the deploy

CU-869cj5zpj


packages/cli/src/commands/verifyManifest.ts
```diff
@@ -0,0 +1,61 @@
...
+
+        try {
+            const result = await verifyManifestOnServer(
+                httpClient,
+                accessToken,
+                resolvedAppType,
+                manifestContent as unknown as Record<string, unknown>,
+            );
+
+            if (result.valid) {
```
silviojaeger:
```suggestion
            if (result.data.valid) {
```

packages/cli/src/utils/verifyManifest.ts
```diff
@@ -238,3 +240,31 @@ export const verifyManifest = (manifest: unknown, schema: typeof platformAppMani
 
     return validatedManifest.data;
 };
+
+export type VerifyManifestResult = { valid: true } | { valid: false; error: string };
```
silviojaeger:
```suggestion
export type VerifyManifestResult = {
    success: boolean;
    data: { valid: true } | { valid: false; error: string };
};
```

ragi96: Blindly commit suggestions

packages/cli/src/commands/deploy.ts
```diff
@@ -160,6 +187,10 @@ export const createDeployment = async (
                 ? verifyManifest(manifestContent, platformAppManifestSchemaV1)
                 : manifestContent;
 
+        if (!dryRun && (manifestContent.appType || 'content-block') === 'content-block') {
```
mike85: @ragi96 would it make sense to also run validation in `dryRun` mode? Since it’s only a validation, there’s no downside, but skipping it makes `dryRun` less reliable.
Right now it can give a green light even though the real deployment would fail, which is misleading. If I use `dryRun`, I want it to reflect reality as closely as possible. Wdyt?

ragi96: true

packages/cli/src/commands/verifyManifest.ts
```diff
@@ -0,0 +1,61 @@
...
+                Logger.success('The manifest is valid.');
+                return;
+            }
+
+            Logger.error('The manifest is invalid:', result.data.error);
+            process.exit(-1);
+        } catch (error) {
+            Logger.error(
+                'An error occurred while verifying the manifest:',
+                (error as HttpClientError).responseBody?.error || (error as Error).message,
```
Kenny806:
```suggestion
                (error as HttpClientError).responseBody?.error ?? (error as Error).message,
```

packages/cli/src/commands/deploy.ts
Kenny806: Why not prefer `??`?
```suggestion
        if (!dryRun && (manifestContent.appType ?? 'content-block') === 'content-block') {
```
Also, the fallback to `'content-block'` when `manifestContent.appType` is undefined/null raises some questions. Is it a good idea to assume something is a content block? Not sure how exactly these manifests are structured, so maybe I just need some clarification.

packages/cli/src/commands/verifyManifest.ts
```diff
@@ -0,0 +1,61 @@
...
+    token?: string;
+};
+
+export const verifyManifest = async ({ appType, instance, token }: VerifyManifestOptions): Promise<void> => {
+    try {
+        const { instanceUrl, accessToken } = resolveCredentials(token, instance);
+
+        const projectPath = process.cwd();
+        const manifestContent = reactiveJson<AppManifest>(join(projectPath, 'manifest.json'));
+        const resolvedAppType = appType || manifestContent.appType || 'content-block';
```
Kenny806:
```suggestion
        const resolvedAppType = appType ?? manifestContent.appType ?? 'content-block';
```
And same question about the `'content-block'` fallback. Does that make sense?

packages/cli/src/commands/deploy.ts
ragi96: It's basically coming from the first iteration of the cli (blocks only), and handled like this everywhere, let's remove these fallbacks for v7 so it's handled everywhere the same, wdyt?

ragi96: adjusted

Kenny806 (approved):

mike85 (approved):
LGTM

fulopdaniel (approved):

julianiff (approved):

PR #1568, opened 2026-04-28
feat(sidebar-settings): extend sidebar settings structure to provide more flexibility

Kenny806:
This PR introduces new `SettingsSection` and `TranslatableSettingsStructure` in the `sidebar-settings` package. 

The goal is to enable creating theme/block settings in a way, which makes them (optionally) translatable and extendable on the level of settings section, not just on the level of individual setting blocks. 
A concrete example of such extendibility on settings section level is adding an optional icon to the section's title. Rather than relying on using the section's camelCased key, splitting it into words and capitalizing their first letter.

The current settings type `{ [customSectionName: string]: SettingBlock[] };` will still be compatible, compatibility will be handled internally by web-app. 

This PR also includes an update to the `guideline-themes` package, so that the new settings type can be used by themes.

### DIFF
--- a/packages/sidebar-settings/src/settingsStructure.ts
+++ b/packages/sidebar-settings/src/settingsStructure.ts
@@ -55,9 +55,16 @@ export type SettingsSection<AppBridge> = MainSection<AppBridge> | LabeledSection
 /**
  * Settings structure that resolves at render time for the requested language.
  *
+ * The argument is an object so the contract can grow new fields in the future
+ * without breaking existing callers.
+ *
  * @typeParam AppBridge - The AppBridge variant the contained blocks target.
  *
  * Currently supported languages are:
  * 'en', 'de', 'fr', 'es', 'it', 'pt', 'nl', 'da', 'sv', 'pl', 'cs', 'fi', 'no'
  */
-export type TranslatableSettingsStructure<AppBridge> = (lang: string) => SettingsSection<AppBridge>[];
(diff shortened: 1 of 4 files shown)


packages/sidebar-settings/src/settingsStructure.ts
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Human-readable section label. Provided by the consuming theme/block,
+     * typically translated via the {@link TranslatableSettingsStructure}
+     * function form.
+     */
+    label: string;
+    /**
+     * Optional leading icon as a React element.
+     */
+    icon?: ReactElement;
```
Kenny806, line 36: Decided to drop support for `IconEnum` here. There are only two ways to support it, neither of which is clean:

1. The blocks way: use the deprecated `DynamicFondueIcon` component and `LegacyIconEnum` 🚫 
2. Write a custom function that would translate `IconEnum` into a `@frontify/fondue/icons` element: too much overhead 🚫 

As the blocks use deprecated symbols, they will eventually be forced to either drop support or implement a new solution, at which point the icon support will either be in sync or we can adopt the new solution. Having `ReactElement` only makes this more future-proof and reduces the chance of having to deal with breaking changes.

ragi96: Yep imo the way to go 👍

mike85 (approved):
LGTM

ragi96 (approved):
LGTM

PR #1571, opened 2026-04-30
chore(useAttachment): fix test so it works with v4 and v3

ragi96:
the three failing tests now wait on content (attachments length 3) instead of reference identity.


Kenny806 (approved):
🚀

PR #1572, opened 2026-04-30
refactor: remove translatability from theme settings structure

Kenny806:
To simplify things, we decided to not make settings translatable at this point and only allow defining an icon and a label for sections


ragi96 (approved):
Dont forget the changeset :)

mike85 (approved):

PR #1599, opened 2026-06-10
feat(app-bridge-theme): add support for hiding navigation pages

Kenny806:


.changeset/witty-points-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/app-bridge-theme": patch
+---
+
+Extend DocumentNavigationItem and DocumentPageNavigationItem to enable working with pages hidden from navigation
```
ragi96: Please scope the changeset as well, It makes them so much nicer.

You can use the `/create-changeset` skill which we added in this repo

packages/app-bridge-theme/src/types/Guideline.ts
```diff
@@ -89,6 +90,7 @@ interface DocumentPageBase {
 }
 export interface DocumentPageNavigationItem extends DocumentPageBase {
     headings(): DocumentPageHeadingNavigationItem[];
+    isHiddenInNavigation(): boolean;
```
anxobotana: Per our conversation, do not expose this unless it is strictly needed

anxobotana (changes requested):

.changeset/witty-points-tap.md
Kenny806: Thanks for the suggestion. Hopefully it is better now. BTW, I had to add the scope manually - the skill did not include it

ragi96: Thanks for the feedback, will have a look :)

anxobotana (approved):
LGTM

PR #1621, opened 2026-07-14
fix(BlockItemWrapper): prevent showing the tooltip when dragging by mouse

peter-tudosa:
Small tweak in logic to prevent showing tooltip when dragging by mouse.


ragi96 (approved):
LGTM

PR #1623, opened 2026-07-14
chore!: remove DownloadButton component

peter-tudosa:
It looks like this component is not used anywhere, we even have an exact duplicate in `guideline-blocks` repo.


ragi96 (commented):
Can we create a workpackage to collect all things we wanna remove for v3? 
And then do it altogether as soon we have a bigger bag of it? Seems not that senseful to do a new major release only for removing that button

PR #1626, opened 2026-07-21
feat(Attachments): several qol changes

fulopdaniel:
The attachments component needed quite some love so here is the list of improvements in this PR:

1. Loading indicator is now shown when selecting asset via asset chooser.
2. Hover color is now less prominent (using surface tokens)
3. On mobile it is properly taking full width now
4. Using Fondue `<Button />` components instead of native buttons
5. Migrate to the new `AssetInput` component
6. Focus visible outlines no longer clipped
7. DnD now stays inside the flyout content

https://github.com/user-attachments/assets/67ac2990-04ea-40fe-baba-bb8bfe4ae4d1

### DIFF
--- a/.changeset/tricky-pants-grab.md
+++ b/.changeset/tricky-pants-grab.md
@@ -1,0 +1,5 @@
+---
+"@frontify/guideline-blocks-settings": patch
+---
+
+feat(Attachments): several qol changes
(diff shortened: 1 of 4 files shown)


peter-tudosa (approved):
Good work, don't forget the changeset! 🚀

PR #1628, opened 2026-07-22
fix(cli):  make eslint usage optional during deploy

peter-tudosa:
The deploy command should not crash/stop if it cannot find an `eslint` config file.

We're migrating `guideline-blocks` packages to `oxlint` and the current deploy setup does not work anymore.
The current changes will fix the issue making the deploy script a bit more flexible.

P.S: current changes will require the consumers to run a `pnpm lint` command before the deploy manually (if they are not using eslint) but as discussed within the team we think this should be the way to go as deployment script should not be responsible for linting.

### DIFF
--- a/.changeset/eager-webs-drop.md
+++ b/.changeset/eager-webs-drop.md
@@ -1,0 +1,5 @@
+---
+"@frontify/frontify-cli": patch
+---
+
+fix(deploy): prevent script from crashing if package does not use eslint
--- a/.changeset/eager-webs-drop.md
+++ b/.changeset/eager-webs-drop.md
@@ -2,4 +2,6 @@
 "@frontify/frontify-cli": patch
 ---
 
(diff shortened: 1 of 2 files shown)


.changeset/eager-webs-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@frontify/frontify-cli": patch
+---
+
+fix(deploy): prevent script from crashing if package does not use eslint
```
ragi96:
```suggestion
fix(deploy): skip eslint checks when no eslint config is present

The deploy command previously failed when a project had no ESLint config file. It now detects a missing config and skips the ESLint step, so projects that lint with another tool (e.g. oxlint) can deploy. Run your  linter manually before deploying if you don't use ESLint.
```

ragi96 (approved):
LGTM but would improve the changeset

julianiff (approved):

PR #1640, opened 2026-08-13
feat(app-bridge): add link chooser related commands

peter-tudosa:
## What this PR adds

Adds `openLinkChooser`/`closeLinkChooser` block commands, a `linkChosen` event, and a `useLinkChooser` hook — mirroring the existing `openAssetChooser`/`assetsChosen` pattern to drive the in-sourced link chooser.

## Usage

```ts
import { openLinkChooser, closeLinkChooser } from '@frontify/app-bridge';

await appBridge.dispatch(openLinkChooser({ selectedUrl: currentUrl }));
// ...
await appBridge.dispatch(closeLinkChooser());
```
Or via the hook:

```ts 
const { openLinkChooser, closeLinkChooser } = useLinkChooser(appBridge);
await openLinkChooser((url) => setUrl(url), { selectedUrl: 'xxx' });
````

### DIFF
--- a/packages/app-bridge/src/react/useLinkChooser.ts
+++ b/packages/app-bridge/src/react/useLinkChooser.ts
@@ -1,0 +1,30 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+import { type EventUnsubscribeFunction } from '../AppBridge';
+import { type AppBridgeBlock } from '../AppBridgeBlock';
+import { closeLinkChooser, openLinkChooser } from '../registries/commands/LinkChooser';
+import { type LinkChooserOptions } from '../types';
+
+type UseLinkChooserType = {
+    openLinkChooser: (callback: (url: string) => void, options: LinkChooserOptions) => Promise<void>;
+    closeLinkChooser: () => Promise<void>;
+};
+
+// oxlint-disable-next-line @eslint-react/no-unnecessary-use-prefix
+export const useLinkChooser = (appBridge: AppBridgeBlock): UseLinkChooserType => {
+    let unsubscribe: EventUnsubscribeFunction;
+
+    return {
+        openLinkChooser: async (callback, options) => {
+            unsubscribe?.();
+            unsubscribe = appBridge.subscribe('linkChosen', (chosenLink) => {
+                callback(chosenLink.url);
+            });
+            await appBridge.dispatch(openLinkChooser(options));
+        },
+        closeLinkChooser: async () => {
+            unsubscribe?.();
+            await appBridge.dispatch(closeLinkChooser());
+        },
+    };
+};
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: packages/app-bridge/src/react/useLinkChooser.ts [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+import { type AppBridgeBlock } from '../AppBridgeBlock';
+import { closeLinkChooser, openLinkChooser } from '../registries/commands/LinkChooser';
+import { type LinkChooserOptions } from '../types';
+
+type UseLinkChooserType = {
+    openLinkChooser: (callback: (url: string) => void, options: LinkChooserOptions) => Promise<void>;
+    closeLinkChooser: () => Promise<void>;
+};
+
+// oxlint-disable-next-line @eslint-react/no-unnecessary-use-prefix
```
fulopdaniel, line 13: **question**: is this ox-lint disable needed? seems to work without it.

packages/app-bridge/src/types/LinkChooserOptions.ts
```diff
@@ -0,0 +1,6 @@
+/* (c) Copyright Frontify Ltd., all rights reserved. */
+
+export type LinkChooserOptions = {
+    hubId: number;
```
fulopdaniel: **naming suggestion:** maybe `portalId` to match what we will [pass the API](https://github.com/Frontify/app-server/pull/33100)?

fulopdaniel: **thought:** on another note, I wonder if it makes sense to omit this parameter alltogether, and just always use `currentPortalId` in web-app. I mean e.g. opening it via the `LinkPlugin`, how is `brand-sdk/packages/guideline-blocks-settings` supposed to know about the current hub id?

packages/app-bridge/src/types/LinkChooserOptions.ts
```diff
@@ -0,0 +1,6 @@
...
+    selectedItem?: string;
```
fulopdaniel: **naming nit (non-blocking):** `selectedUrl`

packages/app-bridge/src/react/useLinkChooser.ts
peter-tudosa: Didn't work for me, because basically this is not a react hook, there's no react stuff going on in here, but I kept the pattern that the other commands use.

packages/app-bridge/src/types/LinkChooserOptions.ts
peter-tudosa: Good point! Took it out!

packages/app-bridge/src/types/LinkChooserOptions.ts
peter-tudosa: changed!

fulopdaniel (approved):

PR #1664, opened 2026-09-04
feat(app-bridge): add openBrandChecker command

mpetrone:
### What does this PR do?

Adds the `openBrandChecker` command to `@frontify/app-bridge`.

A content block can dispatch this command to open the Brand Check dialog on a guideline page. The payload `{ tab?: 'check' | 'configuration' }` is optional. The host opens the check tab when the payload is missing.

The web-app host supplies `guidelineProjectId` and `brandId` at mount time. The command does not take those IDs.

### How to test

1. Install this package in a content block.
2. Call `appBridge.dispatch(openBrandChecker())` or `appBridge.dispatch(openBrandChecker({ tab: 'configuration' }))`.
3. Confirm TypeScript accepts the call. The web-app handler lands in a follow-up after this alpha is published.

### Related notes

Draft for the content-and-blocks team. Web-app PR: https://github.com/Frontify/web-app/pull/16648

### DIFF
--- a/.changeset/open-brand-check.md
+++ b/.changeset/open-brand-check.md
@@ -1,0 +1,7 @@
+---
+'@frontify/app-bridge': minor
+---
+
+feat: add `openBrandCheck` command
+
+A content block can dispatch `openBrandCheck` to open the Brand Check dialog on a guideline page. The payload `{ tab?: 'check' | 'configuration' }` is optional. The host opens the check tab when the payload is missing.
--- a/.changeset/open-brand-check.md
+++ b/.changeset/open-brand-check.md
@@ -2,6 +2,6 @@
 '@frontify/app-bridge': minor
 ---
 
-feat: add `openBrandCheck` command
+feat: add `openBrandChecker` command
 
-A content block can dispatch `openBrandCheck` to open the Brand Check dialog on a guideline page. The payload `{ tab?: 'check' | 'configuration' }` is optional. The host opens the check tab when the payload is missing.
+A content block can dispatch `openBrandChecker` to open the Brand Check dialog on a guideline page. The payload `{ tab?: 'check' | 'configuration' }` is optional. The host opens the check tab when the payload is missing.
--- a/.changeset/open-brand-check.md
+++ b/.changeset/open-brand-check.md
@@ -4,4 +4,4 @@
 
 feat: add `openBrandChecker` command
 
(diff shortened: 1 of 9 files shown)


packages/app-bridge/src/registries/commands/CommandRegistry.ts
```diff
@@ -21,6 +21,7 @@ type DownloadAsset = Asset;
 type OpenSearchDialog = void;
 type CloseSearchDialog = void;
 type OpenPlatformAppDirect = { marketplaceAppId: string };
+export type OpenBrandCheckPayload = { tab?: 'check' | 'configuration' };
```
peter-tudosa: Would you mind moving this to the types file so we keep consistency, also is the tab optional on purpose? is there a situation where you call this function without specifying the tab?

.changeset/open-brand-check.md
```diff
@@ -0,0 +1,7 @@
+---
+'@frontify/app-bridge': minor
```
peter-tudosa: Not sure if it impacts something along the way (I am thinking about scripts that gather up changelogs), but to be on the safe side because this is how all the other changeset files do it.
```suggestion
"@frontify/app-bridge": minor
```

packages/app-bridge/src/registries/commands/CommandRegistry.ts
mpetrone: types added. We will need it because in the design we will be able to open different tabs.

peter-tudosa: Exactly so I don't think it should be optional that's what I am saying.
As far as I can tell from the description you can call `appBridge.dispatch(openBrandChecker())` or `appBridge.dispatch(openBrandChecker({ tab: 'configuration' }))` but you never do `appBridge.dispatch(openBrandChecker({}))`, right?

mpetrone: ahh i see, i misunderstood it. The 'check' behavior is the default behavior but i can force the parameter and make the web-app to always send it.

.changeset/open-brand-check.md
peter-tudosa:
```suggestion
"@frontify/app-bridge": minor
```

peter-tudosa (approved):

PR #1669, opened 2026-09-07
fix(RichTextEditor): assert zIndex auto on the floating link and button portals

peter-tudosa:
## Problem

The floating link and button modals are portaled to `document.body` and rely on DOM order so that a dialog opened from them — notably the guideline LinkChooser — renders on top. Opening the chooser from an RTE link left it rendering *behind* the modal that opened it.

Root cause is not a z-index we set. In production that portal computes **`z-index: 1`**, and DOM order actually favoured the dialog (modal at body index 37, dialog at 38) — the `1` overrode it. Ruled out as the source:

- `plate-floating` — emits only `position/top/left/display/visibility`
- `BlockStyles[TextStyles.p]` — fonts/margins/colour only
- all 47 bundled `@frontify` stylesheets, plus every stylesheet in web-app, guideline-blocks and guideline-themes — zero rules whose selector can match the portal (it is class-less and id-less)

The remaining candidate is **guideline portal custom CSS** setting a `z-index` on body children, confirmed not to use `!important`. That is not something a consumer can be relied upon not to do.

## Fix

Assert `zIndex: 'auto'` inline on all four portals (link insert/edit, button insert/edit). Inline styles win over those rules, so ordering is restored without this package introducing a stacking value of its own.

Each file carries a comment explaining why the value is there, so it does not get tidied away as a no-op.

## Notes for review

- Companion: Frontify/web-app#16679 gives the same treatment to web-app's own copy of these plugins, which had a hardcoded `zIndex: 50`.
- This deliberately avoids adding a z-index layer to fondue's `Dialog`, per review feedback on Frontify/fondue#2844 that portaled elements should rely on DOM order.
- **Known gap:** this only helps blocks rebuilt and redeployed against a released SDK. Blocks already live in customer guidelines keep the `z-index: 1` and stay broken. A defensive layer in fondue's `Dialog` would have covered those immediately — worth a decision from FPT rather than leaving it implicit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/guideline-blocks-settings/src/components/RichTextEditor/plugins/ButtonPlugin/components/FloatingButton/CustomFloatingButton.tsx
+++ b/packages/guideline-blocks-settings/src/components/RichTextEditor/plugins/ButtonPlugin/components/FloatingButton/CustomFloatingButton.tsx
@@ -33,6 +33,13 @@ export const CustomFloatingButton = () => {
     const input = <InsertButtonModal />;
     const editContent = isEditing ? input : <EditModal />;
 
+    /*
+     * `zIndex: 'auto'` is asserted deliberately: these modals are portaled to `document.body`
+     * and must stay ordered by DOM position, so a dialog opened from them (e.g. the guideline
+     * LinkChooser) renders on top. Guideline portals can carry custom CSS that sets a z-index
+     * on body children, which would otherwise lift this above that dialog. Inline styles win
+     * over those rules, so do not remove it.
+     */
     return (
         <>
             {isOpen &&
@@ -42,7 +49,7 @@ export const CustomFloatingButton = () => {
                         data-is-underlay
                         ref={insertRef}
                         {...insertProps}
-                        style={{ ...insertProps.style, ...BlockStyles[TextStyles.p] }}
+                        style={{ ...insertProps.style, ...BlockStyles[TextStyles.p], zIndex: 'auto' }}
                     >
                         {input}
                     </div>,
@@ -56,7 +63,7 @@ export const CustomFloatingButton = () => {
                         data-is-underlay
                         ref={editRef}
                         {...editProps}
-                        style={{ ...editProps.style, ...BlockStyles[TextStyles.p] }}
+                        style={{ ...editProps.style, ...BlockStyles[TextStyles.p], zIndex: 'auto' }}
                     >
                         {editContent}
                     </div>,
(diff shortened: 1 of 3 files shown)


peter-tudosa:
Retargeted from `main` to `dev` (rebased, diff unchanged).

Both consumers are on the 4.0.0-alpha line, which is published from `dev` — so a `main`-based fix would have shipped in a 3.0.x release that neither uses:

| Consumer | Specifier | Resolved |
| --- | --- | --- |
| guideline-blocks (all 20 block packages) | `4.0.0-alpha.2` (exact pin) | `4.0.0-alpha.2` |
| web-app | `^4.0.0-alpha.2` | `4.0.0-alpha.2` |

Two follow-ups worth a decision:

- **guideline-blocks pins the version exactly**, so it needs a bump PR to whatever alpha carries this before any block picks it up. web-app's caret will float on its own.
- **The 3.0.x line does not get this fix.** If marketplace blocks are still built against 3.x, this wants a companion PR against `main` — happy to open one if that line is still live.

packages/guideline-blocks-settings/src/components/RichTextEditor/plugins/LinkPlugin/FloatingLink/CustomFloatingLink.tsx
```diff
@@ -57,6 +57,13 @@ export const CustomFloatingLink = () => {
         <EditModal editButtonProps={editButtonProps} unlinkButtonProps={unlinkButtonProps} />
     );
 
+    /*
+     * `zIndex: 'auto'` is asserted deliberately: these modals are portaled to `document.body`
+     * and must stay ordered by DOM position, so a dialog opened from them (e.g. the guideline
+     * LinkChooser) renders on top. Guideline portals can carry custom CSS that sets a z-index
+     * on body children, which would otherwise lift this above that dialog. Inline styles win
+     * over those rules, so do not remove it.
+     */
```
peter-tudosa:
```suggestion

```

packages/guideline-blocks-settings/src/components/RichTextEditor/plugins/ButtonPlugin/components/FloatingButton/CustomFloatingButton.tsx
```diff
@@ -33,6 +33,13 @@ export const CustomFloatingButton = () => {
     const input = <InsertButtonModal />;
     const editContent = isEditing ? input : <EditModal />;
 
+    /*
+     * `zIndex: 'auto'` is asserted deliberately: these modals are portaled to `document.body`
+     * and must stay ordered by DOM position, so a dialog opened from them (e.g. the guideline
+     * LinkChooser) renders on top. Guideline portals can carry custom CSS that sets a z-index
+     * on body children, which would otherwise lift this above that dialog. Inline styles win
+     * over those rules, so do not remove it.
+     */
```
peter-tudosa:
```suggestion

```

ragi96 (approved):

